PR #4, opened 2016-09-07
Biggish Refactor

freenerd:
I integrated with `osrm-frontend` and refactored a bunch of things to make that smoother.

@bsudekum can u have a look?

- Make `compile` sycnhronous for easier integration into
  leaflet-routing-machine
- Use `throw` as error method in `compile`
- Use `v5` instead of `5` as version flag (to make it more semantic)
- Move fixtures into own files so they can be easier re-used to test
  non-JS implementation in the future


bsudekum:
This is great, was thinking of doing the same thing to the tests here. 

Looks good.

PR #9, opened 2016-09-07
Handle unknown turn types

freenerd:
> new identifiers might be introduced without API change Types unknown to the client should be handled like the turn type, the existance of correct modifier values is guranteed.

https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md#properties-4

/cc @bsudekum


bsudekum:
Looks good. One thought- we could take a closer look at the modifier and make a more educated guess as to what the instruction should like. IE, if modifier is `straight`, we could say instead assign the `type` to `continue`.

PR #11, opened 2016-09-07
Add all the tests [WIP]

bsudekum:
Adds a test file for every modifier/type combination.

Hopefully this will not be a pain to maintain.

/cc @freenerd


freenerd:
Good. Stuff.

> Hopefully this will not be a pain to maintain.

Only really annoying thing that we need to manually maintain are file name changes. I think this will be okay. Love that this finally gives us a complete picture of all possible text instructions 😄

bsudekum:
I did add an update=true npm test feature that will auto update all fixtures btw.

Issue #16, opened 2016-09-09
Better U-turn instructions


bsudekum:
Although there are some combinations that I doubt would happen, we should still cover all cases with understandable instructions. example https://github.com/Project-OSRM/osrm-text-instructions/blob/master/test/fixtures/v5/fork/uturn.json

```json
{
    "step": {
        "maneuver": {
            "modifier": "uturn",
            "type": "fork"
        },
        "name": "Street Name"
    },
    "instruction": "Keep uturn at the fork onto Street Name"
}
```

PR #17, opened 2016-09-09
Cleanup a few obscure instructions

bsudekum:
Part 1 to https://github.com/Project-OSRM/osrm-text-instructions/issues/16

As it stands some obscure modifier/type combinations have very strange instructions. This fixes a few of these cases.

/cc @freenerd

### DIFF
--- a/instructions.json
+++ b/instructions.json
@@ -18,7 +18,9 @@
             "defaultInstruction": "Head {modifier}[ on {way_name}]"
         },
         "end of road": {
-            "defaultInstruction": "Turn {modifier}[ onto {way_name}]"
+            "defaultInstruction": "Turn {modifier}[ onto {way_name}] at the end of the road",
+            "straight": "Continue straight[ onto {way_name}]",
+            "uturn": "Make a U-turn[ onto {way_name}] at the end of the road"
         },
         "ferry": {
             "defaultInstruction": "Take the ferry[ - {way_name}]"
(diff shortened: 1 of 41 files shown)


instructions.json
```diff
@@ -18,25 +18,35 @@
             "defaultInstruction": "Head {modifier}[ on {way_name}]"
         },
         "end of road": {
-            "defaultInstruction": "Turn {modifier}[ onto {way_name}]"
+            "defaultInstruction": "Turn {modifier}[ onto {way_name}] at the end of the road",
```
freenerd: I personally don't like `at the end of the road` because it is really long. I'd imagine `end of road` could be used for a special turn symbol, but should not find it's way into text instructions.

/cc @1ec5 @MoKob

bsudekum: Yeah I was thinking the same thing while making this change actually

MoKob: Well. It is helpful if you have a long way to the intersection. So I feel what I would prefer is to have a long version for announcing what comes next and a short version that would omit this additional information at the turn.

So if you make the turn, you could get an immediate information `at the end of the road, turn left`. And when you are at the intersection `turn left`. This at the end-of-the-road would replace the distance in the announcement.

I imagine: `Turn left at the end of the road` instead of `In 200 meters, turn left at the end of the road`.

bsudekum: > I imagine: Turn left at the end of the road instead of In 200 meters, turn left at the end of the road.

@MoKob correct, however this repo will not have any idea about distances. When this module is used by the developer, it's their responsibility to add the distances.

MoKob: Well, if its only about minimising the wording in a restricted setting, then using `turn` is probably the safer bet. I have so say though, that I fear we might limit ourselves here if this repo is going to be the main point to get instructions from. I might have a conflicting impression of what we want to achieve here, though.

freenerd: >  this repo will not have any idea about distances

I don't think we should bring distances per se into this repo, but we could include alarm priorities with a placeholder for the client to integrate the distance. Likely this would be an override as we have already, so only special cases (like end of road high priority) would get a custom text.

MoKob: Wouldn't it be low priority (https://github.com/mapbox/directions/blob/master/user-guidance.md#step-terminology) which would replace the distance? In addition, the user would need to know about our preference not to add the distance in that case. After all, end of road provides all you need to know in terms of when to take the turn.

bsudekum: Per chat, I think adding an optional distance string is a good way forward. Adding alert priority level adds some overhead that this repo should not deal with.

bsudekum:
Going to keep `end of the road` verbose for now. This can be revisited when be break up instructions to have two different types (verbose, non-verbose).

Issue #8, opened 2016-09-07
Never allow english strings outside of instructions.json


bsudekum:
Can we lint for this?

/cc @freenerd

bsudekum:
First step was done in https://github.com/Project-OSRM/osrm-text-instructions/pull/20

Next, rotary logic needs to have a few english phrases removed

freenerd:
I think #27 should have fixed this completely, everything is in instructions.json.

PR #20, opened 2016-09-14
Remove english from lane logic

bsudekum:
per https://github.com/Project-OSRM/osrm-text-instructions/issues/8, the only english strings should be found in the english translation file. This removes english from the lanes js file and attempts to simplify each lane configuration down to a mappable string.

/cc @1ec5 @freenerd


instructions.json
```diff
@@ -93,7 +93,27 @@
             "straight": "Go straight[ onto {way_name}]"
         },
         "use lane": {
-            "defaultInstruction": "Continue {modifier}"
+            "defaultInstruction": "{laneInstruction} and continue straight[ onto {way_name}]",
+            "left": "{laneInstruction} and turn left[ onto {way_name}]",
```
1ec5: The conjunction here should be "to", not "and". "And" when spoken tends to make the user think there are two different steps, and that would get even more confusion when linking instructions together.

instructions.json
```diff
@@ -93,7 +93,27 @@
...
+            "right": "{laneInstruction} and turn right[ onto {way_name}]",
+            "sharp left": "{laneInstruction} and make a sharp left[ onto {way_name}]",
+            "sharp right": "{laneInstruction} and make a sharp right[ onto {way_name}]",
+            "slight left": "{laneInstruction} and make a slight left[ onto {way_name}]",
+            "slight right": "{laneInstruction} and make a slight right[ onto {way_name}]",
+            "uturn": "{laneInstruction} and make a U-turn[ onto {way_name}]",
+            "laneTypes": {
+                "o": "Take any lane",
+                "oo": "Take any lane",
```
1ec5: If the user can take any lane, why say so?

Issue #7, opened 2016-09-07
handle waypoints


bsudekum:
ref https://github.com/Project-OSRM/osrm-text-instructions/blob/f1d966c1589926e84f36c79e6d62682f3d8510ea/index.js#L38

freenerd:
Reason for implementing this is that we used to issue the following `arrive` instructions, in cases of multiple legs on a route.

- `You have arrived at your second destination, on the right`
- `You have arrived at your final destination, on the right`

I don't feel that counting strongly that this is a useful feature. I think instead the implementing application needs to figure out, how several legs are displayed to the outside of text instructions.

1ec5:
Fixed in #111.

PR #21, opened 2016-09-20
Depart/Arrive dont always have modifiers

freenerd:
From https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md

> The list of turns without a modifier is limited to: depart/arrive. If the source/target location is close enough to the depart/arrive location, no modifier will be given.

/cc @bsudekum

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,8 +36,10 @@ module.exports = function(_version) {
             // This switch statement is for specical cases that occur at runtime
             switch (type) {
             case 'arrive':
-                // TODO, add wayPoint argument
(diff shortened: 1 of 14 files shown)


freenerd:
Added another change with disabling the `{nth}` Waypoint for now. It's better to not have this, than to keep exposing through the variable. Tracked in https://github.com/Project-OSRM/osrm-text-instructions/issues/7 I also changed all arrive instructions to only say `to the {left|right}` instead of `shard/slight` because I feel these don't add much semantic anyways, given that cars are in motion during arrival and it still takes some destination finding anyways.

PR #23, opened 2016-09-21
Support destinations

### LINKED ISSUE #22: Use destinations
We should include destinations in the text instructions.

It should say `Take the ramp toward Baltimore`.

We might be able to tack this onto the `{way_name}`, but then `toward` would not be localizable.

/cc @bsudekum

freenerd:
Fixes #22 

Also adds `way_name` to `off ramp` and `on ramp`. I assume that in most cases these ramps will either have no `name` or `destinations` instead.

/cc @bsudekum @1ec5 @MoKob


MoKob:
With 5.4, we should see mostly destinations on ramps. `names` can happen, though, since the ramp is often named the same way as the motorway.

Should we possibly prefer destinations over names for ramps?

freenerd:
@MoKob that is exactly what is happening, ignoring `name` if `destinations` is present.

Compare these two:

https://github.com/Project-OSRM/osrm-text-instructions/blob/69e2da43ad22fe5efbea201d7886dd9cd1cb455d/test/fixtures/v5/on_ramp/slight_right_destinations.json
https://github.com/Project-OSRM/osrm-text-instructions/blob/69e2da43ad22fe5efbea201d7886dd9cd1cb455d/test/fixtures/v5/on_ramp/slight_right.json

freenerd:
I'm merging this quickly so it can go out on osrm-frontend https://github.com/Project-OSRM/osrm-frontend/pull/187.

Still, please feel free to comment, we can fix in new PRs

Issue #6, opened 2016-09-07
Can we reduce special logic in index.js even more?


bsudekum:
The smaller this logic is, the more extensible this library will be. We should try to see if we can reduce this logic even further.

/cc @freenerd

1ec5:
#23 muddies up the format even further: now the `[optional]` syntax can be replaced outright with a `destination` "template". For localization purposes, that PR should've created a separate set of messages for each towards case instead of that "template".

freenerd:
When I implemented #23 my natural way to go would have actually been named `[]` variables. That seems complex though and ripe for weird string manipulation cases (e.g. setting and retaining spaces correctly). Instead, we could also remove the `[]` variables completely, by splitting every `turn/modifier` combination into a new object with several cases.

Now:

```
        "off ramp": {
            "defaultInstruction": "Take the ramp[ onto {way_name}]",
         }
```

New:

```
        "off ramp": {
            "defaultInstruction":  {
                 "default": "Take the ramp",
                 "name": "Take the ramp onto {way_name}",
                 "destination": "Take the ramp towards {destination_name}",
            }
         }
```

Issue #24, opened 2016-09-21
Roundabout text instruction does not replace name correctly


freenerd:
http://map.project-osrm.org/?z=12&center=50.517902%2C4.766350&loc=50.504256%2C4.792099&loc=50.565794%2C4.827461&hl=en&alt=0

> Enter the roundabout and take the second exit onto {way_name}

freenerd:
Fixed in #27 

> Enter the roundabout and take the second exit towards Bruxelles

Issue #19, opened 2016-09-12
Make instructions.json format localizable


1ec5:
For the instructions strings file, we should use a format that standard localization tools can work with. What we have now is more or less the [JSON key-value format](http://docs.transifex.com/formats/json/) supported by Transifex, but with an informal mini-language around placeholders and optional clauses. This format has some downsides for localization:
1. No way to provide comments to localizers
   - Switching to [Chrome JSON](http://docs.transifex.com/formats/chrome_json/) format would address this issue.
2. No way to vary capitalization or word order when a clause is omitted
   - We should get rid of all `[optional]` clauses and enumerate all the possibilities. String concatenation always works against localizability.
3. No way to vary surrounding words based on a placeholder’s number or grammatical gender
   - We could build this ourselves by allowing any string to be replaced by an object mapping number patterns (`ONE`, `TWO`, `MANY`) to message variants, but localizers would be unable to add their own variants using a tool like Transifex.

Primarily for (3) and so that other clients don’t have to build their own custom parsers, we should consider switching to one of the [other Transifex formats](http://docs.transifex.com/formats/) that happens to support these features. As an iOS developer, I’m most familiar with [.strings](http://docs.transifex.com/formats/apple-strings/) and .stringsdict, but [gettext](http://docs.transifex.com/formats/gettext/) and [XLIFF](http://docs.transifex.com/formats/xliff/) are solid formats that support all the features we need for localization while being compatible [with Node](https://www.npmjs.com/package/grunt-xliff), [Android](https://developer.android.com/distribute/tools/localization-checklist.html), and [iOS](https://developer.apple.com/library/ios/documentation/MacOSX/Conceptual/BPInternational/LocalizingYourApp/LocalizingYourApp.html) development.

Depends on #6 and #5.

/cc @bsudekum @MoKob @freenerd @zugaldia

freenerd:
After #27 this is were we are at:

> 1. No way to provide comments to localizers

This is not there yet, using [chrome json](http://docs.transifex.com/formats/chrome_json/) as suggested feels like a good idea. Need to do a sample implementation of that.

> 2. No way to vary capitalization or word order when a clause is omitted
> We should get rid of all [optional] clauses and enumerate all the possibilities. String concatenation always works against localizability.

We have no optional clauses anymore, everything is spelled out completely in whole strings. We still have string interpolation via `{token}`.

> 3. No way to vary surrounding words based on a placeholder’s number

[...]

Issue #14, opened 2016-09-08
Roundabouts/Rotaries should include exit numbers and street name


freenerd:
First, roundabouts should have the same handling as rotaries.

Then, currently roundabouts only cover the first case, but instead should cover all. Let's stay in line with #8 and have all of the string in the instructions.json. Maybe this is the point where we'd need a better templateing language, that can support nesting of attributes, so that we could have a string like this:

`Enter {rotary_name} [=nameandexit and take the {exit} exit onto {way_name}] [=name and exit onto {way_name}] [=exit and take the {exit} exit]`

```
    case 'rotary':
        instruction = instruction.replace('{rotary_name}', step.rotary_name || 'the rotary');
        if (step.name && m.exit) {
            instruction = instruction + ' and take the {exit} exit onto {way_name}';
        } else if (step.name) {
            instruction = instruction + ' and exit onto {way_name}';
        } else if (m.exit) {
            instruction = instruction + ' and take the {exit} exit';
        }
        break;
```

freenerd:
Added roundabout handling being the same as rotary handling in https://github.com/Project-OSRM/osrm-text-instructions/pull/13/commits/9df6319fe57ad5da55958d892cd080993e3c2928 https://github.com/Project-OSRM/osrm-text-instructions/pull/13

Big issue remains #8 here

freenerd:
This has been fixed in #27 with [nested properties](https://github.com/Project-OSRM/osrm-text-instructions/blob/1eae991700536a657004b440640b1c5f352a1f4e/instructions.json#L293-L316) `name`, `exit` and `name_exit`.

PR #27, opened 2016-10-05
Refactor for more data-driven approach

freenerd:
Improves https://github.com/Project-OSRM/osrm-text-instructions/pull/23
Fixes https://github.com/Project-OSRM/osrm-text-instructions/issues/6
Ref https://github.com/Project-OSRM/osrm-text-instructions/issues/19  https://github.com/Project-OSRM/osrm-text-instructions/issues/24 https://github.com/Project-OSRM/osrm-text-instructions/issues/14 https://github.com/Project-OSRM/osrm-text-instructions/issues/8

This makes natural language translations possible and implementations in other programming languages easier on the cost of blowing up the instructions.json.

Next:

- [x] @bsudekum @1ec5 can you have a look?
- [x] Fix tests for rotary/roundabout (we need to incorporate the nesting here as well)
- [x] 🎉


index.js
```diff
@@ -20,84 +20,73 @@ module.exports = function(_version) {
...
+                if (!laneInstruction) {
                     // If the lane combination is not found, default to continue
-                    instruction = instructions[version][type].defaultInstruction;
+                    instructionObject = instructions[version][type].default;
                 }
                 break;
-            default:
+            case 'rotary':
+            case 'roundabout':
+                if (type === 'rotary' || type === 'roundabout') {
```
1ec5: This if statement is redundant to the cases we're in.

index.js
```diff
@@ -20,84 +20,73 @@ module.exports = function(_version) {
...
+            var nthWaypoint = ''; // TODO, add correct waypoint counting
             instruction = instruction
-                // If a modifier is not provided, calculate direction given bearing
-                .replace('{modifier}', modifier || utils.getDirectionFromDegree(step.maneuver.bearing_after)[0])
-                .trim();
+                .replace('{nth}', nthWaypoint)
+                .replace('{destination}', (step.destinations || '').split(',')[0])
+                .replace('{exit_number}', ordinalize(step.maneuver.exit || 1))
+                .replace('{rotary_name}', step.rotary_name)
+                .replace('{laneInstruction}', laneInstruction)
```
1ec5: Most of the tokens use train_case, but this one uses camelCase.

instructions.json
```diff
@@ -1,111 +1,381 @@
...
-            "slight left": "{laneInstruction} to make a slight left[ onto {way_name}]",
-            "slight right": "{laneInstruction} to make a slight right[ onto {way_name}]",
-            "uturn": "{laneInstruction} to make a U-turn[ onto {way_name}]",
+            "default": {
+                "default": "Continue straight",
+                "name": "Continue straight onto {way_name}",
+                "destination": "Continue straight towards {destination}"
+            },
+            "left": {
+                "default": "{laneInstruction} to turn left",
```
1ec5: For the purpose of minimizing translation cost, it may be desirable at some point to refactor this so that there would literally be a string `{laneInstruction} to {instruction}`. I'm not aware of any languages that would need more granularity in this case, but I could be wrong.

instructions.json
```diff
@@ -1,111 +1,381 @@
...
+            },
+            "sharp left": {
+                "default": "Take a sharp left",
+                "name": "Take a sharp left onto {way_name}",
+                "destination": "Take a sharp left towards {destination}"
+            },
+            "sharp right": {
+                "default": "Take a sharp right",
+                "name": "Take a sharp right onto {way_name}",
+                "destination": "Take a sharp right towards {way_name}"
```
bsudekum: Should this be destination?

instructions.json
```diff
@@ -1,111 +1,381 @@
...
+                "default": "Make a U-turn",
+                "name": "Make a U-turn onto {way_name}",
+                "destination": "Make a U-turn towards {destination}"
+            }
         },
         "notification": {
-            "defaultInstruction": "Continue {modifier}[ onto {way_name}]"
+            "default": {
+                "default": "Continue {modifier}",
+                "name": "Continue {modifier} onto {way_name}"
```
bsudekum: Is it impossible to have a destination here?

bsudekum:
Some other thoughts:
* we should add a test that verifies that each type has the 3 required keys
* Should the key `name` be renamed to `way_name` so it corresponds to the template we're replacing? `{way_name}`? Downside here is that `_` in keys is a little icky looking to me, upside is it's more clear. 

I think there are some good moves here going on but with this PR we're 3x'ing the instruction.json and only reducing `index.js` by a couple of lines.

freenerd:
Incorporated all inline comments in https://github.com/Project-OSRM/osrm-text-instructions/pull/27/commits/f54c828403d69d77fb34e3319a8fd61298b04c35

> we should add a test that verifies that each type has the 3 required keys

Yes. It feels like the test fixtures need to eventually look similar to the instructions.json. I might also want to generate them programatically then. Specifically lanes, rotary and roundabouts will need work.

> Should the key name be renamed to way_name so it corresponds to the template we're replacing?

👍

1ec5:
> we're 3x'ing the instruction.json and only reducing index.js by a couple of lines.

In general, that's to be expected with making anything more data-driven, although I've suggested a change above that would reduce the string bloat a bit.

freenerd:
Some more changes:

- Automatically generate fixtures https://github.com/Project-OSRM/osrm-text-instructions/pull/27/commits/ccf59509ae79aa249850e4bf3fb2843eff87152c
- Constants in instructions.json https://github.com/Project-OSRM/osrm-text-instructions/pull/27/commits/a61ca5ee1e5d431f6969121b564d67b2bf31958f https://github.com/Project-OSRM/osrm-text-instructions/pull/27/commits/c12c332474332c640ad2c55b04d0dad3c8cbfe4f https://github.com/Project-OSRM/osrm-text-instructions/pull/27/commits/be80c99e18f701a675b600795c672925a59a12ec

instructions.json
```diff
@@ -1,112 +1,407 @@
...
+                "destination": "{lane_instruction} to turn right towards {way_name}"
+            },
+            "uturn": {
+                "default": "{lane_instruction} to make a U-turn",
+                "name": "{lane_instruction} to make a U-turn onto {way_name}",
+                "destination": "{lane_instruction} to make a U-turn towards {destination}"
+            },
+            "lane_types": {
                 "xo": "Keep right",
                 "ox": "Keep left",
```
bsudekum: We should move these to `.constants`

1ec5:
On iOS, I was planning to use the Mapbox iOS/macOS SDK's MGLCompassDirectionFormatter class for compass directions (which would be localized along with the SDK), but this is a reasonable approach.

freenerd:
Consolidated `use lane` as well.

I think we are good here now, next step is merging, releasing a new version, integrating and then continue to fix things. Will do that tomorrow.

Issue #30, opened 2016-10-18
Use most common/popular ref and destination


freenerd:
Ways can have several `ref` and `destinations`. These are usually `;` delimited. Currently we are only always using the first one. Instead, we should use the most common or popular one.

As an example, imagine a cross-country road that has the following refs on the way:
- `I-80`
- `I-80;I-81`
- `I-80;I-81;DETOUR-20`
- `I-80`
- `I-90;I-80`
- `I-90;I-80`
- `I-80`

It would be great if the user would always only get `I-80` as the ref.

We could implement this via a pre-processing step where we create a priority list (based on occurence, maybe mixed with length of staying on segments) and when deciding for a ref within a way, we should use the one with highest priority.

freenerd:
During the pre-processing step, we may also handle mode changes https://github.com/Project-OSRM/osrm-text-instructions/issues/18 and waypoints https://github.com/Project-OSRM/osrm-text-instructions/issues/7

freenerd:
Note that it likely makes sense to have one combined priority list for destinations and ref, since [OSRM is mixing](https://github.com/Project-OSRM/osrm-backend/blob/3f0f0e306b3c2ee5eada6512c3ce3ab00b1b0460/profiles/lib/destination.lua#L7) the `destination:ref` tags into the `destinations` field

TheMarex:
Cross-linking https://github.com/Project-OSRM/osrm-text-instructions/issues/115#issue-240502408 here as another test case for this.

freenerd:
There is a prototypical implementation in https://github.com/Project-OSRM/osrm-text-instructions/pull/77

lyzidiamond:
Per convo with @1ec5 and @mcwhittemore, there are two separate-but-related issues at play here:

1. When on a multiplexed road like `I-80;I-90`, instructions should pick the better one for the route. Considerations include: the previous instruction (if the user is already traveling on `I-80` they should continue on it), the subsequent instruction (if the user is traveling on `I-80` but will be continuing onto `I-90` after they split), the instructions on the majority of the route (if most of the route travels on `I-80`, regardless of immediately preceding and subsequent instructions).
[...]

PR #32, opened 2016-10-18
Add handling for name and ref

### LINKED ISSUE #29: Use ref and name together
Currently `step.ref` is ignored, but shouldn't be.

freenerd:
Fixes #29 

This also introduces automatically-generated fixtures for `type: turn`


test/fixtures/v5/turn/left_ref_default.json
```diff
@@ -0,0 +1,11 @@
+{
+    "step": {
+        "maneuver": {
+            "type": "turn",
+            "modifier": "left"
+        },
+        "name": "",
+        "ref": "Ref1;Ref2"
+    },
+    "instruction": "Turn left onto Ref1"
```
freenerd: @bsudekum @1ec5 @willwhite @karenzshea highlighting here that I've changed the behaviour to only include the first `ref`, instead of all. This is a debatable decision, but I value conciseness highly here, thus one ref should be sufficient. Choosing the best ref should be smarter though, e.g. via #30.

index.js
```diff
@@ -109,12 +109,24 @@ module.exports = function(_version) {
                 // NOOP, since no special logic for that type
             }
 
+            // Decide way_name with special handling for name and ref
+            var wayName;
+            var name = (step.name || '').replace(' (' + step.ref + ')', ''); // Remove hack from Mapbox Directions mixing ref into name
```
1ec5: Use a regular expression to make sure we only replace this substring at the end of the name.

PR #35, opened 2016-10-18
Only mention end-of-the-road on uturn

freenerd:
Let's only mention end of the road when it makes semantic a lot of change, otherwise keep it short.

We might want to revise this decision, e.g. by moving end of the road to the front of the instruction. For now I'm keeping parity with the state we had before in Mapbox Directions.


freenerd:
I think we should keep them separate for now, since me may still choose to include "end of road" information, like "At the end of the road, turn left ...".

PR #38, opened 2016-10-18
No way_name/destination with use lane

freenerd:
I'm a bit unclear how `type: use lane` will actually be used.

It feels to me that `name` and `destination` only make sense, if they actually distinguish the lanes from each other. Since all lanes are on the same OSM way though, all lanes would have the same `name/destination` information. Thus we should not include that information in the text instruction, since it will be the same for all lanes.

We could re-introduce `destination` instructions if OSRM would support the [`destination:lanes` OSM tag](https://wiki.openstreetmap.org/wiki/Lanes), ticketed here https://github.com/Project-OSRM/osrm-backend/issues/2553#issuecomment-227067935 already

/cc @MoKob @daniel-j-h @bsudekum

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file. For change 
 # Unreleased
 
 - Better u-turn instructions
+- Do not use name/destination on use lane
 - Fix rotary names instruction `exit onto`
 - Only mention end-of-road on uturn intructions
 
--- a/instructions.json
+++ b/instructions.json
@@ -36,8 +36,8 @@
             "lanes": {
                 "xo": "Keep right",
                 "ox": "Keep left",
-                "xox": "Use the middle lane",
-                "oxo": "Use the left or right lane"
+                "xox": "Keep in the middle",
+                "oxo": "Keep left or right"
             }
         },
         "arrive": {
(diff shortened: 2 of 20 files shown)


daniel-j-h:
UseLane will be emitted e.g. in Lane Anticipation: take [this file](https://github.com/Project-OSRM/osrm-backend/blob/master/features/guidance/anticipate-lanes.feature) as an example (search for "use lane").

Where you could take multiple lanes going straight but we constraint it to specific lanes, you'll get a UseLane with constraint lane data in the response.

This has no influence on the destination value, since a destination tag without lane namespace applies to the overall way and not to specific lanes.

The only way to find out: sample OSM for destination tags right in front of turns that have lanes attached.

freenerd:
> The only way to find out: sample OSM for destination tags right in front of turns that have lanes attached.

This is out-of-scope for `osrm-text-instructions`. Should I add that as a new issue on `osrm-backend`?

@daniel-j-h I take from your comment that the implementation in this PR makes sense?

daniel-j-h:
I looked deeper into this, turns out there is a problem in OSRM during collapsing: the way we collapse steps could result in us emitting destination signs for (collapsed) steps later.

Ticketed in https://github.com/Project-OSRM/osrm-backend/pull/3097.

freenerd:
Merging this and will fix https://github.com/Project-OSRM/osrm-text-instructions/issues/42 later

PR #40, opened 2016-10-19
Add localization mechanism along with german translation

contargo-development:
This adds a simple localization mechanism for instructions. The instructions JSON files are now suffixed with a language identifier, e.g. `instructions_en.json`. The correct json is retrieved by the instructions module.

In order to use osrm-text-instructions in osrm-frontend, a few changes to `itinerary_builder.js` and `index.js` files are necessary to propagate the language identifier from osrm-frontend to osrm-text-instructions. I can file another PR in osrm-frontend with these changes if desired.


index.js
```diff
@@ -1,9 +1,13 @@
...
+    var language = _language || 'en';
 
-if (Object !== instructions.constructor) throw 'instructions must be object';
+    var instructions = require('./instructions').get(language);
+    if (Object !== instructions.constructor) throw 'instructions must be object';
 
-module.exports = function(_version) {
-    var version = _version || 'v5';
+    function capitalizeFirstLetter(string) {
+        return string.charAt(0).toUpperCase() + string.slice(1);
```
1ec5: This function isn't effective or desirable for all languages. Moreover, the overall instruction string may need to be placed in running text or title-cased depending on the context. Could capitalizing the first letter be a responsibility of client code rather than this library?

1ec5: See also https://github.com/Project-OSRM/osrm-text-instructions/issues/19#issuecomment-254682885.

1ec5:
See also https://github.com/Project-OSRM/osrm-text-instructions/issues/19#issuecomment-254682885.

index.js
contargo-development: If it's depending on the context, I agree that captilizing should be handled by client code. But if we do that, shouldn't all translated strings be lowercase?

contargo-development:
@1ec5 There weren't any problems in translating the string to German. Sometimes, I had to change the location of the interpolated part, e.g. sometimes {modifier}s have to be at the start of the strings to make sense in German. This lead to some instructions starting with a lowercase letter, hence the capitalization function above.

freenerd:
Changes here are awesome. Thanks so much for the translation.

I've rebased on `master` and pushed to [this branch](https://github.com/Project-OSRM/osrm-text-instructions/commits/localization), with a couple of more changes:

- Moved translations into an own directory https://github.com/Project-OSRM/osrm-text-instructions/commit/4b8382eabeae8770b000463369f095150da92e17
- Make languages always need to be defined and add tests https://github.com/Project-OSRM/osrm-text-instructions/commit/8fabed40cfe36c0d46b21d26511cbab7f412b52d
- Put capitalization behind a flag in the translated file https://github.com/Project-OSRM/osrm-text-instructions/commit/5452313acca6fa67396ed28cf65befd53265ef9e

There is some more work to do to make sure that the German and English instructions files have the exact same top-level ordering. Ideally we'd have tests that compare structure.

Some differences I noticed while looking at the `npm test` output with the german language:

```
not ok 543 use_lane/oox.json
  ---
    operator: equal
    expected: 'Keep left to turn left'
    actual:   'Links halten um links abzubiegen auf Street Name'
    at: Test.<anonymous> (/Users/johan/Code/mapbox/osrm-text-directions/test/index_test.js:218:34)
  ...
not ok 544 use_lane/oox_destinations.json
  ---
    operator: equal
    expected: 'Keep left to turn left'
    actual:   'Links halten um links abzubiegen Richtung City 1'
    at: Test.<anonymous> (/Users/johan/Code/mapbox/osrm-text-directions/test/index_test.js:218:34)
  ...
not ok 545 use_lane/right.json
  ---
    operator: equal
    expected: 'Continue straight'
    actual:   'Geradeaus weiterfahren auf Street Name'
    at: Test.<anonymous> (/Users/johan/Code/mapbox/osrm-text-directions/test/index_test.js:218:34)
```

These should not have street names and destinations as per https://github.com/Project-OSRM/osrm-text-instructions/pull/38

```
    expected: |-
      'Make a U-turn towards Destination 1'
    actual: |-
      '180°-Wendung abbiegen auf Destination 1'
```

Maybe a better translation would be `Wenden auf Destination 1`?

```
not ok 523 turn/sharp_right_destination.json
  ---
    operator: equal
    expected: 'Make a sharp right towards Destination 1'
    actual:   'Scharf rechts abbiegen auf Destination 1'
```

`Scharf rechts abbiegen Richtung Destination 1`?

```
    expected: 'Take the ramp on the right onto Street Name'
    actual:   'Rampe auf der rechten Seite nehmen auf Street Name'
```

Shall we re-name `Rampe` as `Ausfahrt` and `Auffahrt`? Not sure if that is semantically correct /cc @MoKob  

@contargo-development Do you want to pull these changes into your your own branch and update there or shall I take it from here in another PR on this repo?

freenerd:
I added some tests for the structure of the language files https://github.com/Project-OSRM/osrm-text-instructions/commit/f2268233f59997098f21643592f9e229629c4857 and removed the un-used types `train` and `pushing bike` (mirroring https://github.com/Project-OSRM/osrm-text-instructions/pull/46)

freenerd:
I fixe the `use lane ` behaviour as described above in https://github.com/Project-OSRM/osrm-text-instructions/commit/f7765d4068a72c2a06445349318b2d8d5f355bc3.

To move this forward, I will close this PR tomorrow and open a new one, so we can get this into master. Goal is to get this into [osrm-frontend](https://github.com/Project-OSRM/osrm-frontend) then 😄

rkcpi:
@freenerd Go ahead! I'd be happy to see this in osrm-frontend :-)

freenerd:
Moving to #58

PR #45, opened 2016-10-19
Switch back to literal matching for ref/name mapbox hack

freenerd:
Reverts https://github.com/Project-OSRM/osrm-text-instructions/pull/32#discussion_r84092116 which broke because of missing tests. 

To use regular expressions, we'd need to sanitize the ref coming in correctly. That feels cumbersome, and given that we'll remove this hack in the near future again I don't think we should invest too much resources into it.


index.js
```diff
@@ -113,7 +113,7 @@ module.exports = function(_version) {
 
             // Decide way_name with special handling for name and ref
             var wayName;
-            var name = (step.name || '').replace(new RegExp(' (' + step.ref + ')$'), ''); // Remove hack from Mapbox Directions mixing ref into name
+            var name = (step.name || '').replace(' (' + step.ref + ')', ''); // Remove hack from Mapbox Directions mixing ref into name
```
1ec5: FYI, we could conditionalize the replacement based on whether the string ends in the parenthesized ref.

Issue #18, opened 2016-09-12
Handling mode changes


freenerd:
Currently modes are incorrectly modeled as turn types. This is not how OSRM returns these.

I believe travel modes are fairly static in OSRM, so these could become part of the instructions.json

- https://github.com/Project-OSRM/osrm-backend/blob/c306a5985476b4008885296b1abd02912097c161/include/extractor/travel_mode.hpp#L42-L54
- https://github.com/Project-OSRM/osrm-backend/blob/c306a5985476b4008885296b1abd02912097c161/profiles/bicycle.lua#L383-L384

What I'm not clear is if we can avoid having access to the previous step to calculate if something changed. Might need to have previous and current 😢

freenerd:
A first step has been implemented in #46, but only for `mode: ferry` and with strong assumptions around that. To get better at this and handle more modes, we'd need to hit #43 to not emit too many mode change instructions.

1ec5:
As mentioned above, `ferry` is currently implemented, but `walking` (for dismounting and walking a bicycle) is unimplemented. The more straightforward implementation would consist of adding a `walking` property to this object, along with logic to only access this property for cycling directions:

https://github.com/Project-OSRM/osrm-text-instructions/blob/d0e8a960ffc62f35694ba5e9930c56d93ccf879f/languages/translations/en.json#L46-L52

Instead of accepting the previous step as input, `compile` could accept a default mode as an option. For example, the Mapbox Directions API’s `mapbox/cycling` profile would correspond to a `defaultMode` of `cycling`.

mkv123:
I'm planning on working on making osrm-text-instructions be aware of the travel modes and changes of modes. However, as words like "drive" are included in the instructions and would be different for different languages, I'm thinking of adding another layer to the already deep ```instructions[language][version][type][modifier]``` hierarchy.  So the hierarchy would become ```instructions[language][version][mode][type][modifier]```.

As a more backwards compatible alternative with less need for redundant strings (but perhaps a bit more hacky) we could extend the existing "modes" dictionary to contain the strings to be overridden. The existing ones as default/fallback values. So we would have something like ```instructions[language][version].modes[mode].overrides[type][modifier]```.

I'm currently leaning towards the second alternative to reduce the need to re-format every single language js file, but would be very interested in hearing your opinions on this. Do you think alternative 2 is the way to go?

mkv123:
[...]

PR #46, opened 2016-10-19
Handle mode ferry

freenerd:
Ref https://github.com/Project-OSRM/osrm-text-instructions/issues/18

This handles only `mode: 'ferry'` in a very simple way. It does not look at the previous/next step to determine mode changes. This is working under the assumption that a `ferry` is only a single step and that we do not want to announce the modifier, but only that a ferry is happening.

Likely at a later point we'll make this more sophisticated.


freenerd:
Merging this, but still happily taking comments.

PR #47, opened 2016-10-19
Allow runtime customizations of instructions.json

freenerd:
This allows clients to customize instructions.json in-code. I've successfully used this in a project to add html tags like this:

```js
traverse(osrmTextInstructions.instructions).forEach(function (node) {
    if (node.constructor === String) {
        this.update(node.
            replace('{way_name}', '<span class="way-name">{way_name}</span>').
            replace('{destination}', '<span class="destination">{destination}</span>'),
            true
        );
    }
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file. For change 
 - Only mention end-of-road on uturn intructions
 - Respect name and ref for way_name
 - Handle ferry mode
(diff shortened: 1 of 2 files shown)


1ec5:
Wouldn't it be less risky to give client code a callback that accepts a way_name and returns a transformed way_name (in your example, wrapped in a tag)? Now all the replace() calls in this library are a bit less robust in the face of client code that inserts arbitrary text (which could be HTML, Markdown, etc.) into the format string.

PR #77, opened 2016-11-14
Use priority for most common/used name ref

freenerd:
Fixes https://github.com/Project-OSRM/osrm-text-instructions/issues/30

Still very experimental, but wanted to get the PR sooner than later for visiblity.

Todo:

- [ ] Integrate into osrm-frontend to see it first-hand
- [ ] Make implementation airtight
- [ ] Write good extended tests for all cases


freenerd:
I integrated this hacky-ish into [osrm-frontend](https://github.com/Project-OSRM/osrm-frontend/compare/hacky-priority-table-instructions?expand=1). Results so far are promising, see _details_ below

<details>

__left is new, right is new__

[route](http://map.project-osrm.org/?z=12&center=38.899717%2C-77.197151&loc=38.897379%2C-77.094927&loc=38.870120%2C-77.367954&hl=en&alt=0)

This one is interesting since the last part is both on US 29 and US 50, but we prefer US 50 (and don't switch between them within a part anymore, as before)

2016-11-15 at 18 35

[route](http://map.project-osrm.org/?z=10&center=34.036160%2C-117.699280&loc=34.132268%2C-117.224121&loc=33.903476%2C-118.105774&hl=en&alt=0)

We might want to give `ref` a boost over `name`, at least in the beginning CA 210 might have been a better choice.

2016-11-15 at 18 55

[route](http://map.project-osrm.org/?z=11&center=37.683005%2C-122.472153&loc=37.839953%2C-122.298260&loc=37.567984%2C-122.394905&hl=en&alt=0)

2016-11-15 at 18 58

[route](http://map.project-osrm.org/?z=11&center=53.003009%2C13.065491&loc=52.942845%2C13.154411&loc=53.081446%2C13.084030&hl=en&alt=0)

2016-11-15 at 19 00

1ec5:
One pitfall to this approach of omitting the road names along the route is that route markers for surface streets are often infrequent (especially in rural areas in the U.S.) or poorly maintained (especially in urban areas in the U.S.). This is often true of U.S. Routes but increasingly the case as you go down to the state, county, and township level. A driver consuming turn-by-turn guidance for [an exurban/suburban route like this](http://map.project-osrm.org/?z=14&center=39.249371%2C-84.210892&loc=39.211042%2C-84.172869&loc=39.279839%2C-84.248142&hl=en&alt=0) would benefit far more from instructions like:

> Turn right on Paxton Avenue to stay on SR 48

when they’re already on SR 48 than:

> Turn right on SR 48

freenerd:
To avoid above problems, we might want to keep the current decisions between name and destination, but at least choose for either the most common. That would solve the `I-80` problem described in #30

hdaymon:
test

PR #87, opened 2016-11-18
[WIP] Vietnamese

1ec5:
I needed to try my hand at localization to make sure the workflow works. So here’s a Vietnamese localization. Before this PR can land, I need to double-check the test fixtures to make sure everything makes sense. I may need to add a bunch more overrides, because `modifier` only gets prepended with “bên” or “sang” for “left” and “right”, not “straight” and “U-turn”.

/cc @freenerd


freenerd:
This was included in https://github.com/Project-OSRM/osrm-text-instructions/pull/89

PR #90, opened 2016-12-08
Persian Translation

prp-e:
I translated some sentences and phrases of instructions, It will be useful, as there's no good/suitable Persian routing software with services to Iran.

### DIFF
--- a/instructions.js
+++ b/instructions.js
@@ -5,7 +5,7 @@ var instructionsEn = require('./instructions/en.json');
 var instructionsFr = require('./instructions/fr.json');
 var instructionsNl = require('./instructions/nl.json');
 var instructionsZhHans = require('./instructions/zh-Hans.json');
-
+var instructionsFa = require('./instructions/fa.json'); 
 // Table to match tag to required language files
 // A tag can redirect to another tag via the language tag as string value
 var table = {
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Nice, thanks for this!

We're currently preparing the OSRM v5.5 launch. Once this is over @miccolis will have a look here.

1ec5:
@prp-e, note that we're in the process of switching to Transifex for managing this project's translations. If you don't mind, it might be a smoother transition if you upload your translations there as well. See #89 for details.

freenerd:
@prp-e thanks a lot for the translation. I have merged the Transifex translation branch in #89. Sadly that means there are conflicts again in this branch. On the up-side, the persian translation is in Transifex now at https://www.transifex.com/project-osrm/osrm-text-instructions/language/fa/ and can be managed there. I've pulled in the file from your PR there, so you can continue to translate there, and then run through the steps in https://github.com/Project-OSRM/osrm-text-instructions#translations to get this PR up-to-date. Good for you?

freenerd:
As per last comment this should be managed through Transifex now. Once the translation is done there, it can be included here in the same way as Russian and Swedish were in https://github.com/Project-OSRM/osrm-text-instructions/pull/95

PR #92, opened 2016-12-28
Added Russian translation

oxidase:
Added Russian translation. 

`ordinalize` and `direction` are conjugated to have correct phrases `{exit_number} съезд ` and `в {direction} направлении`. If `ordinalize` and `direction` must be conjugated with other noun genders than different translations will be required, like `direction_depart`.


1ec5:
Thanks for the translation, @oxidase! We’re managing translations [in Transifex](https://www.transifex.com/project-osrm/osrm-text-instructions/language/ru/) now. If it isn’t too much trouble, it would help if you could import the translated strings into the project there. Otherwise, I think it should be possible to merge this PR and push it to Transifex.

/cc @freenerd

oxidase:
@1ec5 i have uploaded russian translation to transifex that is better suited for translation reviews. My translation is more verbose than it can be, so probably a review is required to shorten sentences.

PR #95, opened 2017-02-23
Add Russian, Swedish

1ec5:
Added complete Russian and Swedish translations from Transifex.

/cc @frederoni @oxidase


frederoni:
Swedish translation LGTM

PR #98, opened 2017-04-02
Update Russian translations from Transifex

1ec5:
Running `npm run transifex` produced these changes in the Russian translation, which means the repository is out of sync with Transifex. @yuryleb, can you review? Thanks!

### DIFF
--- a/languages/translations/ru.json
+++ b/languages/translations/ru.json
@@ -112,8 +112,8 @@
         "end of road": {
             "default": {
                 "default": "Поверните {modifier}",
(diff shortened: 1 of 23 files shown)


yuryleb:
Yes, I'm preparing yet another Russian translation update but it's not completely checked yet. I'll add new PR with these changes soon.

1ec5:
Ah, that makes sense. I’ll merge this now, but feel free to keep working on the Russian updates – no rush. 👍

PR #99, opened 2017-04-02
Spanish

1ec5:
Added Spanish translation from Transifex.

@josek5494, thank you for your contributions! Please review the test fixtures in this PR for any edge cases that may need to be overridden. Incidentally, it looks like we got very lucky in that we only ended up needing feminine ordinals (for _destina_ and _salida_).

/cc @MallorcaRedes

### DIFF
--- a/languages.js
+++ b/languages.js
@@ -2,6 +2,7 @@
 // with bundling tools like webpack and browserify
 var instructionsDe = require('./languages/translations/de.json');
 var instructionsEn = require('./languages/translations/en.json');
+var instructionsEs = require('./languages/translations/es.json');
 var instructionsFr = require('./languages/translations/fr.json');
 var instructionsNl = require('./languages/translations/nl.json');
 var instructionsRu = require('./languages/translations/ru.json');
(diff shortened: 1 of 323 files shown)


1ec5:
Also, if any strings need to differ from country to country (such as Spain versus Colombia), please let me know and I’ll be happy to add the appropriate languages to the Transifex project.

yuryleb:
Please don't forget to remove Spanish code replacement in frontend's [itinerary_builder.js](https://github.com/Project-OSRM/osrm-frontend/blob/gh-pages/src/itinerary_builder.js#L5) code after all.

1ec5:
Oh, interesting. @freenerd, what’s the story behind that special case? Do we need to remove special cases anywhere else in the frontend?

perliedman:
@1ec5 my guess is that the frontend relies on existing i18n in Leaflet Routing Machine for languages that are not yet in osrm-text-instructions.

yuryleb:
But my guess is easier - frontend has its own [Spanish translation block](https://github.com/Project-OSRM/osrm-frontend/blob/gh-pages/i18n/es.js) - this automatically enables Spanish 'Change language' option - so to prevent 'Missing language' error from osrm-text-instructions, this lang code is temporary replaced :wink:

freenerd:
@yuryleb and @perliedman are correct. Ticketed in https://github.com/Project-OSRM/osrm-frontend/issues/232

PR #100, opened 2017-04-04
Don't publish transifex.auth into NPM

yuryleb:
I suppose that transifex.auth file with Transifex authitication was published into NPM repository by mistake. This commit excludes this file from publishing.


freenerd:
Thanks for the find @yuryleb. This was certainly a mistake, I changed my password on transifex.

Issue #101, opened 2017-04-06
Can't extend {way_name} with an option on Transifex


yuryleb:
I prepared new feature for OSRM Text Instructions - grammatical cases support for way names. This is quite important for many languages (Russian, Ukrainian, Polish, Bulgarian, Finnish, Estonian and many others) - otherwise translated strings with inserted way names (in nominative case as in OSM) look grammatically incorrect in most cases. Moreover, I don't know any other navigation system that could do this properly (except my humble attempt to improve [Garmin TTS voices](https://github.com/yuryleb/garmin-russian-tts-voices) :wink:). I would be proud to make OSRM Text Instructions the first system with grammatical case support but I can't :disappointed: 

The main idea of this future feature is in extending `{way_name}` and `{rotary_name}` variables in target language's strings with explicit grammatical case label - `{way_name:accusative}` for example - and processing way names with block of regular expressions grouped by this grammatical case.

But [Trasifex](https://www.transifex.com/project-osrm/osrm-text-instructions/) rejects my attempts to change _"Turn left onto `{way_name}`"_, in Russian _"Поверните налево на `{way_name}`"_ to _"Поверните налево на `{way_name:accusative}`"_ with "Missing `{way_name}` variable error" :disappointed_relieved: Other variants (`{way_name}{accusative}` for example) also don't work due to difference with variables from original English string.

My experience with TTS voices reveals that determining required grammatical case for part of whole text is not the simple task and could be easily done only for known and expected words combinations. Actually this could be implemented too but extends already big regular expressions file (already ~900 expressions for Russian street names only) some more.

I don't like the extending all translation strings with proposed grammatical case option - this has no sense for them. I don't want to extend English strings only with this option - this invalidates all existing translations. I would like to update Russian (currently) translation only. Maybe some [Transifex checks](https://docs.transifex.com/setup/setting-translation-checks#section-custom-variables) could be suppressed/updated to allow non-exact match of variables?

@1ec5, @freenerd, @TheMarex, need your advice please.

yuryleb:
Whole feature description for reference (from uncommitted changes):

## Grammar support

Many languages - all Slavic (Russian, Ukrainian, Polish, Bulgarian, etc), Finnic (Finnish, Estonian) and others - have [grammatical case feature](https://en.wikipedia.org/wiki/Grammatical_case) that could be supported in OSRM Text Instructions too.
[...]

Issue #53, opened 2016-10-21
Swift port


freenerd:
Documenting that a swift port will happen:
- [ ] Transform instruction files to `.plist`
- [ ] Execute integration test suite
- [ ] Build and package

/cc @1ec5 @bsudekum

1ec5:
To be clear, this would consist of a script that could convert the JSON file to a plist on demand, plus whatever minimal Swift code we need to match the functionality currently in index.js.

freenerd:
I'm not too-well-versed with the Apple ecosystem, but should this maybe be Objective C instead of Swift to reach full compatibility with all apps (new and old)?

1ec5:
Swift can bridge to Objective-C to some extent. But you're right, we might want to use Objective-C for the benefit of applications that need static linking. We'd end up using the same Foundation APIs either way.

1ec5:
Here’s how I’d prefer the Swift support to be structured:

* osrm-text-instructions implements an Instruction struct that holds all the data relevant to an instruction about a step. It would look similar to MapboxDirections.swift’s RouteStep class, minus all the Objective-C compatibility gunk.
* osrm-text-instructions implements an InstructionFormatter subclass of Formatter that takes an Instruction instance and outputs a localized string formatted for display or voice.
* MapboxDirections.swift implements a RouteStepFormatter subclass of Formatter that internally converts a RouteStep into an Instructions instance in order to then convert it into a user-friendly instruction string.

MapboxDirections.swift will depend on osrm-text-instructions, not the other way around. This way, MapboxDirections.swift can remain focused on the Mapbox Directions API without making compromises for direct OSRM-frontend usage. Meanwhile, osrm-text-instructions is shielded from the Objective-C API compatibility concerns that currently make MapboxDirections.swift a pain to develop. (Since osrm-text-instructions is such a new library, I don’t think there’s any demand for Objective-C compatibility for it. But if the demand arises, we can always write an ObjC compatibility layer in Swift.)

1ec5:
[OSRM Text Instructions for Swift](https://github.com/Project-OSRM/osrm-text-instructions.swift/) is functional and available.

Issue #26, opened 2016-10-04
Ruby version


zavan:
I didn't know how to contact you guys, so I'm creating this issue, I hope that's not a problem.

Just so you guys know, I know it's very soon but I've ported this lib almost verbatin to a ruby gem: https://github.com/zavan/osrm_text_instructions

I plan to keep it up to date with this lib.

Thanks for the great work.

freenerd:
Nice, thanks @zavan. As you said it's still early and we will likely change the data format and logic around a couple of more times, e.g. https://github.com/Project-OSRM/osrm-text-instructions/pull/27

> I plan to keep it up to date with this lib.

🎉

zavan:
No problem, I did it because I really needed it for a project I'm working
on, so It's on my interest to keep it up to date.

Thanks again for your work.
Em qua, 5 de out de 2016 às 11:30, Johan Uhle <notifications@github.com>
escreveu:

freenerd:
@zavan just FYI, a lot has moved in the last month but also the current structure project feels way more robust than before. Also we got translations in. There will like be more breaking changes in the near future, still if you want to bring your port up-to-date, now might be a good time. Make sure you run can execute the test suite, it was very helpful when building a port in Swift for me ...

zavan:
@freenerd The port is necessary for a bigger project of mine, I'm currently working on another part of it, but I'll update the the port as soon as I can, probably late next week. Thanks for the heads up!

Issue #52, opened 2016-10-21
Callback for token replacement values in compile


freenerd:
https://github.com/Project-OSRM/osrm-text-instructions/pull/47 feels brittle, since it depends on the format of the `instructions.json` and therefore couples code tightly.

Instead, @1ec5 proposed to introduce a hook [just before the final token replacement](https://github.com/Project-OSRM/osrm-text-instructions/blob/4d2428224f1835394f21018ccd47219c1f2ae6c9/index.js#L153). That would be a more-stable API, while allowing even more functionality.

1ec5:
Not sure why this was closed. We implemented [a callback](https://github.com/Project-OSRM/osrm-text-instructions.swift/blob/b48a5a072168edd53680c5cb3d86934ef8df9e11/OSRMTextInstructions/OSRMTextInstructions.swift#L179) parameter in the Swift port, and it’s been quite helpful for [marking up the resulting string in SSML](https://github.com/mapbox/mapbox-navigation-ios/blob/b33f7f619305edf1902409dc8629fe734d1f234b/MapboxCoreNavigation/RouteStepFormatter.swift#L24-L35).

1ec5:
The `token()` function would stay as-is, but the `compile()` function would accept an additional `interpolator` function that takes a `tokenType` and `rawValue` and returns a string (typically based on `rawValue`).

1ec5:
If a client wishes to mark up parts of a road name, such as surrounding numbers in an SSML tag as in mapbox/mapbox-navigation-ios#552, then the current approach would be:

1. `way_name` comes from OSRM.
1. The client marks up `way_name` in place.
1. The marked-up `way_name` is passed into `compile()`.
1. The marked-up `way_name` is inserted into the instruction string.

Per https://github.com/Project-OSRM/osrm-text-instructions/pull/102#discussion_r141936188, this approach is too fragile for our needs. This library loses access to the original `way_name`, which is needed for declension logic. The proper order of operations would be:

1. `way_name` comes from OSRM and is passed to `compile()` as-is.
2. `way_name` is declined via the logic in #102.
3. The client is given an opportunity to mark up the declined `way_name` via the callback proposed in this issue.
4. The declined, marked-up `way_name` is finally inserted into the instruction string.

/cc @allierowan @yuryleb

PR #102, opened 2017-04-06
Grammar support for Russian way names

yuryleb:
New feature for OSRM Text Instructions - grammatical cases support for way names. This is quite important for many languages (Russian, Ukrainian, Polish, Bulgarian, Finnish, Estonian and many others) - otherwise translated strings with inserted way names (in nominative case as in OSM) look grammatically incorrect in most cases.

This feature extends {way_name} and {rotary_name} in instructions with optional grammar case label and so requires  update of other applications used OSRM Text Instructions data (Ruby #26 and Swift #53).

Unfortunately this proposal is incompatible with Transifex (see #101) and so can't be easily merged. This PR is supposed mostly for proposed changes review in details and to illustrate this problem for public.


freenerd:
Thanks @yuryleb. As mentioned [before](https://github.com/Project-OSRM/osrm-text-instructions/issues/101#issuecomment-292169554), we need to do some more thinking here.

Blockers seem to be:

- Re-think localization file format https://github.com/Project-OSRM/osrm-text-instructions/issues/19
- Fix workflow with transifex to allow for [different tokens](https://github.com/Project-OSRM/osrm-text-instructions/issues/101#issuecomment-292316003)

yuryleb:
@freenerd, @1ec5, @oxidase, finally I solved to add grammatical case labels in new `languages\overrides\ru.js` script to post-process Transifex localization data - translated strings parsing seems quite simple and could be easily patched for new context phrases or even re-used for other languages if it will be necessary.

I don't think Transifex or any other localization service will allow arbitrary keywords expansion. Actually `languages\overrides` post-processing used by OSRMTI for final customization looks the better idea.

So let's meet the first route instructions with proper grammatical cases at least in Russian :tada:
osrmti-grammar

BTW other OSRMTI-related projects ([osrm-text-instructions.swift](https://github.com/Project-OSRM/osrm-text-instructions.swift), [osrm-text-instructions.java](https://github.com/Project-OSRM/osrm-text-instructions.java), etc?) should also be updated to expect `{way_name:*}` and `{rotary_name:*}` variables at least or completely support grammatical cases (with way described in `Grammar.md`) at last.

Grammar.md
```diff
@@ -0,0 +1,53 @@
+## Grammar support
+
+Many languages - all Slavic (Russian, Ukrainian, Polish, Bulgarian, etc), Finnic (Finnish, Estonian) and others - have [grammatical case feature](https://en.wikipedia.org/wiki/Grammatical_case) that could be supported in OSRM Text Instructions too.
+Originally street names are being inserted into instructions as they're in OSM map - in [nominative case](https://en.wikipedia.org/wiki/Nominative_case).
+To be grammatically correct, street names should be changed according to target language rules and instruction context before insertion.
```
1ec5: Are the various cases always regular and easy to determine from the nominative case? I noticed a few features have been tagged [`name:dative`](https://taginfo.openstreetmap.org/keys/name%3Adative) in OpenStreetMap – would it be desirable or necessary for OSRM to pass such tags along to OSRM Text Instructions?

Grammar.md
```diff
@@ -0,0 +1,53 @@
...
+- __Customization__ - could be easily extended for other languages with adding new regular expressions blocks into [grammar support](languages/grammar/) folder and modifying `{way_name}` and other variables in translated instructions only with necessary grammatical case labels
+
+### Notes
+
+- Russian regular expressions are based on [Garmin Russian TTS voices update](https://github.com/yuryleb/garmin-russian-tts-voices) project; see [file with regular expressions to apply to source text before pronouncing by TTS](https://github.com/yuryleb/garmin-russian-tts-voices/blob/master/src/Pycckuu__Milena%202.10/RULESET.TXT).
+- There is another grammar-supporting module - [jquery.i18n](https://github.com/wikimedia/jquery.i18n) - but unfortunately it has very poor implementation in part of grammatical case applying and is supposed to work with single words only.
+- Actually it would be great to get street names also in target language not from default OSM `name` only - there are several multi-lingual countries supporting several `name:<lang>` names for streets. But this the subject to address to [OSRM engine](https://github.com/Project-OSRM/osrm-backend) first.
```
1ec5: This is a good idea – please file an issue in osrm-backend.

Grammar.md
```diff
@@ -0,0 +1,53 @@
...
+{
+    "v5": {
+        "accusative": [
+            ["^ (\\S+)ая-(\\S+)ая [Уу]лица ", " $1ую-$2ую улицу "],
+            ["^ (\\S+)ая [Уу]лица ", " $1ую улицу "],
+            ...
+```
+- All such JSON files should be registered in common [languages.js](languages.js)
+- Instruction text formatter ([index.js](index.js) in this module) should:
+  - check `{way_name}` and `{rotary_name}` variables for optional grammar case after colon: `{way_name:accusative}`
```
1ec5: I like this syntax – it leaves open the possibility of additional transformations like capitalization, which could replace the "capitalizeFirstLetter" meta property currently set on some localizations.

Grammar.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Design goals
+
+- __Cross platform__ - uses the same data-driven approach as OSRM Text Instructions
+- __Test suite__ - has [prepared test](test/grammar_tests.js) to check available expressions automatically and has easily extendable language-specific names testing pattern
+- __Customization__ - could be easily extended for other languages with adding new regular expressions blocks into [grammar support](languages/grammar/) folder and modifying `{way_name}` and other variables in translated instructions only with necessary grammatical case labels
+
+### Notes
+
+- Russian regular expressions are based on [Garmin Russian TTS voices update](https://github.com/yuryleb/garmin-russian-tts-voices) project; see [file with regular expressions to apply to source text before pronouncing by TTS](https://github.com/yuryleb/garmin-russian-tts-voices/blob/master/src/Pycckuu__Milena%202.10/RULESET.TXT).
```
1ec5: Thank you for [dedicating](https://github.com/yuryleb/garmin-russian-tts-voices/blob/master/LICENSE.txt) this rule set to the public domain!

index.js
```diff
@@ -199,10 +200,36 @@ module.exports = function(version, _options) {
 
             return this.tokenize(language, instruction, replaceTokens);
         },
+        grammarize: function(language, name, grammar) {
+            // Process way/rotary name with applying grammar rules if any
+            if (name && grammar && grammars && grammars[language] && grammars[language][version]) {
+                var rules = grammars[language][version][grammar];
+                if (rules) {
+                    // Pass original name to rules' regular expressions enclosed with spaces for simplier parsing
+                    var n = ' ' + name + ' ';
```
1ec5: It's currently possible for clients to manipulate a step's name before compilation. For example, the Mapbox Directions API currently inserts SSML markup around numbers within the name so that speech synthesizers pronounce them more casually. I think that could potentially interfere with this feature. That would be a good argument for implementing #52, so that clients can reliably manipulate the name after grammaticalization.

Grammar.md
yuryleb: IMHO it's not deal of OSM - it's impossible to store all case variants in all languages for each name. Perhaps (CLDR)[http://cldr.unicode.org/] will be the better place (if will :wink: )

index.js
yuryleb: No problem, if clients will not damage `{way_name:*}` vars - for example, ["Restore names highlighting" PR](https://github.com/Project-OSRM/osrm-frontend/pull/259) adds HTML tags around these vars and all works as on snapshot above.
But if Mapbox Directions inserts SSML tags right instead `{way_name}`, this looks wrong - for example, Nuance uses [`\tn=address\` tags](https://github.com/yuryleb/garmin-russian-tts-voices/blob/master/src/Pycckuu__Milena%202.10/RULESET.TXT#L141) in its Vocalizer voices to specify start of address text specially to pronounce numbers and known street abbreviations as address but doesn't change original address/name text.

yuryleb: But actually it's no problem, even if Mapbox will catch and replace `{way_name}` and skips `{way_name:accusative}` etc - this allows further "grammarization" to work :wink: SSML tags actually important only for English addresses, Russian & many others have no special requirements for them.

1ec5: I’m referring to a JavaScript port of mapbox/mapbox-navigation-ios#552 for the Directions API’s (not yet documented) `voice_instructions` parameter. Essentially we’re working around a problem where Amazon Polly assumes an address is abbreviated, whereas OSM never abbreviates road names. The workaround is to wrap only numbers in `<say-as interpret-as="address">` tags, since we still want numbers to be optimized for speech even if the rest of the words can’t be treated as part of an address. For the JavaScript port, we’re changing the value of `way_name` itself, so that by the time `compile()` runs, `way_name` contains SSML code. 😬

I’ll be the first to admit it’s a hack, but I wanted to point this out because clients may be inclined to do similar things to the value of `way_name` unless we advise against doing so in this project’s documentation. If this PR forces us to implement #52, that’d be a wonderful outcome in my opinion. 😉

index.js
```diff
@@ -199,10 +200,36 @@ module.exports = function(version, _options) {
...
+                    rules.forEach(function(rule) {
+                        var re = new RegExp(rule[0]);
```
1ec5: Should these rules be case insensitive?

test/grammar_test.js
```diff
@@ -0,0 +1,117 @@
...
+        ['Пыхов-Церковный проезд', 'accusative', 'Пыхов-Церковный проезд'],
+        ['Пыхов-Церковный проезд', 'dative', 'Пыхову-Церковному проезду'],
+        ['Пыхов-Церковный проезд', 'genitive', 'Пыхова-Церковного проезда'],
+        ['Пыхов-Церковный проезд', 'prepositional', 'Пыховом-Церковном проезде'],
+        ['Третье Транспортное кольцо', 'accusative', 'Третье Транспортное кольцо'],
+        ['Третье Транспортное кольцо', 'dative', 'Третьему Транспортному кольцу'],
+        ['Третье Транспортное кольцо', 'genitive', 'Третьего Транспортного кольца'],
+        ['Третье Транспортное кольцо', 'prepositional', 'Третьем Транспортном кольце']
+    ]
+    // TODO add your language grammar tests to call grammarize() and check result
```
1ec5: This is what the code below does, right?

index.js
yuryleb: Actually no (and no for `g` option for all replaces not only for the first match), at least for Russian OSM street names this is unnecessary - there are many working street names validators those keep them compliant to [Russian street names agreement](https://wiki.openstreetmap.org/wiki/RU:%D0%92%D0%B8%D0%BA%D0%B8%D0%9F%D1%80%D0%BE%D0%B5%D0%BA%D1%82_%D0%A0%D0%BE%D1%81%D1%81%D0%B8%D1%8F/%D0%A1%D0%BE%D0%B3%D0%BB%D0%B0%D1%88%D0%B5%D0%BD%D0%B8%D0%B5_%D0%BE%D0%B1_%D0%B8%D0%BC%D0%B5%D0%BD%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D0%B8_%D0%B4%D0%BE%D1%80%D0%BE%D0%B3). And also all current grammar regular expressions take into account potentially capital first word letter.

But for general case it looks useful to add some `meta` block into grammar JSON to specify required regular expression options explicitly.

test/grammar_test.js
yuryleb: This is the place to add other language's test strings to check other language's grammar when it will be.

1ec5:
Thanks @yuryleb! I’m going to merge this PR now, but feel free to iterate on the feature in separate PRs if you have any additional changes to make.

Issue #106, opened 2017-05-08
Allow better language file handling


freenerd:
Currently all language files are loaded when osrm-text-instructions is initialized, but only one is used then.

We should instead:

- Allow to control which languages are actually loaded
- Allow switching languages without reloading language files

1ec5:
This is happening in #107.

1ec5:
#107 landed.

PR #107, opened 2017-05-16
Multiple language support

patjouk:
This PR introduces support for multiple language support. ~~Right now all language files are always loaded. With this PR, the developer can decide which language files to require. During test instruction compilation, the language used can be chosen. This allows us to scale this project to large numbers of languages, without bloating data downloaded in browsers for all users.~~

The module API as well as internal APIs have breaking changes, since we had to change several function signatures.

TODOs

- [x] Integrate PR within Mapbox (to see what breaks)
- [x] Integrate PR with [osrm-frontend](https://github.com/Project-OSRM/osrm-frontend/) (to see what breaks)
- [x] Code review
- [x] Update the changelog
- [x] Make sure that relevant docs are updated
- [ ] Merge
- [ ] Release as new version 0.3
- [ ] Update osrm-frontend to use 0.3

CC @freenerd 

edit: We will continue to load every language for now because browserify doesn't support dynamic requires. ([cf @freenerd's comment](ttps://github.com/Project-OSRM/osrm-text-instructions/pull/107#issuecomment-302864645))

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -17,17 +17,18 @@ OSRM Text Instructions has been translated into [several languages](https://gith
 
 ### JavaScript Usage
 
-```
+```js
 var version = 'v5';
-var language = 'en';
-var options = {};
-var osrmTextInstructions = require('osrm-text-instructions')(version, language, options);
+var options = {
+    languages: [ 'en', 'fr' ]
+};
+var osrmTextInstructions = require('osrm-text-instructions')(version, options);
 
-// make your request against the API
+// make your request against the API, save result to response variable
 
 response.legs.forEach(function(leg) {
   leg.steps.forEach(function(step) {
-    instruction = osrmTextInstructions.compile(step)
+    instruction = osrmTextInstructions.compile(step, 'en')
   });
 });
 ```
@@ -35,8 +36,8 @@ response.legs.forEach(function(leg) {
 parameter | required? | values | description
 ---|----|----|---
 `version` | required | `v5` | Major OSRM version
-`language` | required | `en` `de` `zh-Hans` `fr` `nl` `ru` | Language identifier
 `options.hooks.tokenizedInstruction` | optional | `function(instruction)` | A function to change the raw instruction string before tokens are replaced. Useful to inject custom markup for tokens
+`options.languages` | optional | `en` `de` `zh-Hans` `fr` `nl` `ru` | Array of language identifiers that should be supported. Default is loading all language files, which can be huge on websites
 
 ### Development
 #### Architecture
--- a/Readme.md
+++ b/Readme.md
@@ -19,13 +19,11 @@ OSRM Text Instructions has been translated into [several languages](https://gith
 
 ```js
 var version = 'v5';
-var options = {
-    languages: [ 'en', 'fr' ]
-};
-var osrmTextInstructions = require('osrm-text-instructions')(version, options);
+var osrmTextInstructions = require('osrm-text-instructions')(version);
 
 // make your request against the API, save result to response variable
 
+var language = 'en';
 response.legs.forEach(function(leg) {
   leg.steps.forEach(function(step) {
     instruction = osrmTextInstructions.compile(language, step)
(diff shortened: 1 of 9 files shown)


Readme.md
```diff
@@ -69,9 +70,9 @@ To add an own translations:
 
 - Go to [Transifex](https://www.transifex.com/project-osrm/osrm-text-instructions/) and create the new translation there
 - When the translation on Transifex is ready, pull in the translation file:
-  - Create an empty translation file `echo "{}" > languages/translations/{language_tag}.json`
-  - Add the new translation file and language tag to `./languages.js`
-  - If needed: make overrides in `languages/overrides/{language_tag}.json`
+  - Create an empty translation file `echo "{}" > languages/translations/{language_code}.json`
+  - Add the new translation file and language code to `./languages.js`
```
freenerd: > the new translation file

This can be removed, since the code for loading each language file is not present anymore. It's only an array of language codes now that needs to be edited.

test/languages_test.js
```diff
@@ -56,3 +46,12 @@ tape.test('verify language files structure', function(assert) {
 
     assert.end();
 });
+
+tape.test('verify that instructions are only returned for user requested languages', function(assert) {
+    var translations = languages.get(['en', 'fr']);
+
+    assert.deepEqual(Object.keys(translations).sort(), ['fr', 'en'].sort(),
+    'only returns en and fr');
```
freenerd: This line break is a bit weird. I'd prefer either having everything on one line, but since that may not be possible for length reasons I'd prefer one-line-per-parameter:

```js
assert.deepEqual(
  Object.keys(translations).sort(),
  ['fr', 'en'].sort(),
  'only returns en and fr'
);
```

Readme.md
```diff
@@ -17,26 +17,27 @@ OSRM Text Instructions has been translated into [several languages](https://gith
...
   });
 });
 ```
 
 parameter | required? | values | description
 ---|----|----|---
 `version` | required | `v5` | Major OSRM version
-`language` | required | `en` `de` `zh-Hans` `fr` `nl` `ru` | Language identifier
 `options.hooks.tokenizedInstruction` | optional | `function(instruction)` | A function to change the raw instruction string before tokens are replaced. Useful to inject custom markup for tokens
+`options.languages` | optional | `en` `de` `zh-Hans` `fr` `nl` `ru` | Array of language identifiers that should be supported. Default is loading all language files, which can be huge on websites
```
1ec5: We don’t want developers to get the impression that only these six languages are supported, given that we currently support ten ([with more to come](https://www.transifex.com/project-osrm/osrm-text-instructions/languages/)). Perhaps add “etc.” and link to [this folder](https://github.com/Project-OSRM/osrm-text-instructions/tree/master/languages/translations/) for the full list of supported languages?

index.js
```diff
@@ -1,40 +1,54 @@
...
+
+module.exports = function(version, _options) {
     // load instructions
-    var instructions = require('./languages').get(language);
+    var options = {};
+    options.hooks = {};
+    options.hooks.tokenizedInstruction = ((_options || {}).hooks || {}).tokenizedInstruction;
+    options.languages = (_options || {}).languages || languages.supportedCodes;
+
+        // TODO Validate language
```
1ec5: Out of curiosity, what sort of validations do you have in mind?

1ec5:
This PR looks like it’ll address #106.

freenerd:
I tried to integrate these changes into osrm-frontend, but that failed. The reason is browserify, which we use to bundle the `required()` js files. Dynamic requires are not supported https://github.com/substack/node-browserify/issues/377 and there doesn't seem to be a way to specify which files to required upfront.

I'm good with actually walking back on the requirement of https://github.com/Project-OSRM/osrm-text-instructions/issues/106 to select which language files to load, given it makes bundling of js files so hard for browsers. For the server-side, this isn't actually a big problem, since the memory required is marginal. We should definitely keep switching languages during compile though.

Readme.md
patjouk: That's a good idea. I'm changing this.

PR #111, opened 2017-06-09
Add waypoint arrival instruction

bsudekum:
This adds instructions for arriving at waypoints and also fixes an issue with updating fixtures.

Necessary for https://github.com/mapbox/mapbox-navigation-ios/pull/270

/cc @freenerd @patjouk @1ec5


bsudekum:
Note, I had to add a metadata field to the fixture files to configure each compile test to allow for unique legIndex values.

index.js
```diff
@@ -68,7 +68,7 @@ module.exports = function(version, _options) {
 
             return config.join('');
         },
-        compile: function(language, step) {
+        compile: function(language, step, legIndex) {
```
bsudekum: New optional third argument, legInde

1ec5: How would you indicate “final”? −1? (We would also need to add a string for “final” to the translations.)

index.js
```diff
@@ -173,7 +172,7 @@ module.exports = function(version, _options) {
                 .replace('{lane_instruction}', laneInstruction)
                 .replace('{modifier}', instructions[language][version].constants.modifier[modifier])
                 .replace('{direction}', this.directionFromDegree(language, step.maneuver.bearing_after))
-                .replace('{nth}', nthWaypoint)
+                .replace('{nth}', legIndex && legIndex > 0 ? this.ordinalize(language, legIndex) : '')
```
1ec5: According to this logic, if there are multiple destinations, the first destination is announced as, “You have arrived at your destination.”

1ec5: I think it may be necessary to accept yet another parameter, the leg count. (That way we also don’t need to represent the last step index as −1.)

index.js
```diff
@@ -173,7 +175,7 @@ module.exports = function(version, _options) {
                 .replace('{lane_instruction}', laneInstruction)
                 .replace('{modifier}', instructions[language][version].constants.modifier[modifier])
                 .replace('{direction}', this.directionFromDegree(language, step.maneuver.bearing_after))
-                .replace('{nth}', nthWaypoint)
+                .replace('{nth}', waypoint)
```
1ec5: nthWaypoint was a more descriptive name. We aren't dealing with the waypoint itself; we're dealing with an ordinal word representing the one-based leg "index".

index.js
```diff
@@ -162,9 +163,10 @@ module.exports = function(version, _options) {
                 instruction = options.hooks.tokenizedInstruction(instruction);
             }
 
+            var waypoint = legIndex >= 0 && legIndex !== legCount - 1 ? this.ordinalize(language, legIndex + 1) : '';
```
1ec5: If it's the last of several waypoints, should we say "final destination"? The current code says only "destination".

bsudekum: Yes, but we currently do not have a translation for this.

1ec5: This repository must have a source string before translators can provide translations in Transifex. Therefore, assuming that it’s a good idea to say “final destination”, this PR needs to add the “final destination” string to en.json. To avoid breaking the tests, I suppose we’d wait to make use of that string until enough translations are in.

index.js
```diff
@@ -68,10 +68,11 @@ module.exports = function(version, _options) {
 
             return config.join('');
         },
-        compile: function(language, step) {
+        compile: function(language, step, legIndex, legCount) {
```
freenerd: Can we make this optional? For most cases (read: non-multi waypoint cases) we will not need these properties. I'd be good to put them into an options object.

```js
compile: function(language, step, options) {
```

test/fixtures_test.js
```diff
@@ -38,11 +40,17 @@ tape.test('verify existance/update fixtures', function(assert) {
...
+                'legCount': legCount
+            } : {};
+
             fs.writeFileSync(
                 fileName,
                 JSON.stringify({
                     step: step,
-                    instructions: instructionsForLanguages(step)
+                    instructions: instructionsForLanguages(step, legIndex, legCount),
+                    metadata: metadata
```
freenerd: Do we even need to add metadata if it's just an empty object most of the time?

test/fixtures/v5/arrive_waypoint/left.json
```diff
@@ -0,0 +1,25 @@
+{
+    "step": {
+        "maneuver": {
+            "type": "arrive",
+            "modifier": "left"
+        },
+        "name": "Street Name"
+    },
+    "instructions": {
+        "de": "Sie haben Ihr erste Ziel erreicht, es befindet sich links von Ihnen",
```
freenerd: We'll need gendering of the adjective here 😢 Maybe we can get around this with some smart same-gender naming, I'll have a look ...

Now: `Sie haben Ihr erste Ziel erreicht`.
Correct: `Sie haben Ihr erstes Ziel erreicht`.

bsudekum:
@freenerd I think having final makes sense now that we have in between waypoints. Thoughts @1ec5?

1ec5:
Haha, good point about “[final destination](https://en.wikipedia.org/wiki/Final_Destination)”. 🙀 In an ideal world, we’d have a name for each waypoint that we’d say instead of “your destination”, but I suppose “your last destination” is passable.

(The railfan in me would have it say “last scheduled station stop”. 😏)

freenerd:
Readme and Changelog still need fixups.

bsudekum:
@freenerd will update the changelog when I update the package's version.

bsudekum:
@freenerd @1ec5 want to give this another review?

index.js
```diff
@@ -68,7 +68,7 @@ module.exports = function(version, _options) {
 
             return config.join('');
         },
-        compile: function(language, step) {
+        compile: function(language, step, legOptions) {
```
freenerd: should we keep this with `options` only? That would be consistent with the `options` we already have for the module initialization.

test/fixtures/v5/arrive_waypoint_last/uturn.json
```diff
@@ -0,0 +1,25 @@
...
+        "es": "Has llegado a tu destino",
+        "fr": "Vous êtes arrivés à votre destination",
+        "id": "Anda telah tiba di tujuan ke-",
+        "nl": "Je bent gearriveerd op de bestemming.",
+        "ru": "Вы прибыли в пункт назначения",
+        "sv": "Du är framme vid din destination",
+        "vi": "Đến nơi ",
+        "zh-Hans": "您已经到达您的个目的地"
+    },
+    "metadata": {
```
freenerd: this should become `options` then as well

bsudekum:
@freenerd note, while changing `legOptions` -> `options`, I had to change another options variable to `opts`. Can be seen in https://github.com/Project-OSRM/osrm-text-instructions/pull/111/commits/cf6780652fdcdf7d89e8d5496c58e78e49939c9a

1ec5:
> Necessary for mapbox/mapbox-navigation-ios#270

The iOS navigation SDK uses the Swift port of this library for logic such as what’s in this PR. Tracking a Swift port in Project-OSRM/osrm-text-instructions.swift#24.

PR #113, opened 2017-06-19
Changelog and bump - v0.4.0

bsudekum:
/cc @freenerd


freenerd:
Minor version bump, so `v0.4.0`

Issue #114, opened 2017-06-30
Implement Support for Exit Numbers


daniel-j-h:
With https://github.com/Project-OSRM/osrm-backend/pull/4215/files osrm will have a new `exits` property in the `RouteStep` (next to destinations, pronunciations, ...).

We need to adapt the instructions to this new feature.

The primary use-case for exit numbers / exit names is for example the US where exits are numbered sequentially and are crucial for highway navigation. Here is an example:
- https://github.com/Project-OSRM/osrm-backend/blob/2bd7252e82e3a0f8ce854ecf5304c9e731c71d8b/features/guidance/exit-numbers-names.feature#L75-L98
- http://www.openstreetmap.org/way/417524818#map=16/37.3877/-121.9776

Where we should announce taking the ramp similar to:

    Take exit 393 towards Great America Parkway, Bowers Avenue

cc @freenerd @willwhite

1ec5:
Tracking a Swift port in Project-OSRM/osrm-text-instructions.swift#25.

PR #116, opened 2017-07-05
Add support for exits.

willwhite:
This adds support for the [upcoming exits support in OSRM](https://github.com/Project-OSRM/osrm-backend/pull/4215). We are introducing "Take exit {exit}" text here but it's applied pretty narrowly right now - it is only used for off ramps that are tagged with exit information.

Fixes https://github.com/Project-OSRM/osrm-text-instructions/issues/114

### DIFF
--- a/languages/translations/en.json
+++ b/languages/translations/en.json
@@ -254,7 +254,7 @@
                 "name": "Take the ramp on the right onto {way_name}",
                 "destination": "Take the ramp on the right towards {destination}",
                 "exit": "Take exit {exit}",
-                "exit_destination": "Take exit {exit} towards {destination}"
+                "exit_destination": "Take exit {exit} on the right towards {destination}"
             },
             "sharp left": {
                 "default": "Take the ramp on the left",
@@ -268,7 +268,7 @@
                 "name": "Take the ramp on the right onto {way_name}",
                 "destination": "Take the ramp on the right towards {destination}",
                 "exit": "Take exit {exit}",
-                "exit_destination": "Take exit {exit} towards {destination}"
+                "exit_destination": "Take exit {exit} on the right towards {destination}"
             },
             "slight left": {
                 "default": "Take the ramp on the left",
@@ -282,7 +282,7 @@
                 "name": "Take the ramp on the right onto {way_name}",
                 "destination": "Take the ramp on the right towards {destination}",
                 "exit": "Take exit {exit}",
-                "exit_destination": "Take exit {exit} towards {destination}"
+                "exit_destination": "Take exit {exit} on the right towards {destination}"
             }
         },
         "on ramp": {
--- a/languages/translations/en.json
+++ b/languages/translations/en.json
@@ -238,37 +238,51 @@
             "default": {
                 "default": "Take the ramp",
                 "name": "Take the ramp onto {way_name}",
-                "destination": "Take the ramp towards {destination}"
+                "destination": "Take the ramp towards {destination}",
+                "exit": "Take exit {exit}",
+                "exit_destination": "Take exit {exit} towards {destination}"
             },
             "left": {
                 "default": "Take the ramp on the left",
                 "name": "Take the ramp on the left onto {way_name}",
-                "destination": "Take the ramp on the left towards {destination}"
+                "destination": "Take the ramp on the left towards {destination}",
+                "exit": "Take exit {exit} on the left",
+                "exit_destination": "Take exit {exit} on the left towards {destination}"
             },
             "right": {
                 "default": "Take the ramp on the right",
                 "name": "Take the ramp on the right onto {way_name}",
-                "destination": "Take the ramp on the right towards {destination}"
+                "destination": "Take the ramp on the right towards {destination}",
+                "exit": "Take exit {exit}",
+                "exit_destination": "Take exit {exit} towards {destination}"
             },
             "sharp left": {
                 "default": "Take the ramp on the left",
                 "name": "Take the ramp on the left onto {way_name}",
-                "destination": "Take the ramp on the left towards {destination}"
+                "destination": "Take the ramp on the left towards {destination}",
+                "exit": "Take exit {exit} on the left",
+                "exit_destination": "Take exit {exit} on the left towards {destination}"
             },
             "sharp right": {
                 "default": "Take the ramp on the right",
                 "name": "Take the ramp on the right onto {way_name}",
-                "destination": "Take the ramp on the right towards {destination}"
+                "destination": "Take the ramp on the right towards {destination}",
+                "exit": "Take exit {exit}",
+                "exit_destination": "Take exit {exit} towards {destination}"
             },
             "slight left": {
                 "default": "Take the ramp on the left",
                 "name": "Take the ramp on the left onto {way_name}",
-                "destination": "Take the ramp on the left towards {destination}"
+                "destination": "Take the ramp on the left towards {destination}",
+                "exit": "Take exit {exit} on the left",
+                "exit_destination": "Take exit {exit} on the left towards {destination}"
             },
             "slight right": {
                 "default": "Take the ramp on the right",
                 "name": "Take the ramp on the right onto {way_name}",
-                "destination": "Take the ramp on the right towards {destination}"
+                "destination": "Take the ramp on the right towards {destination}",
+                "exit": "Take exit {exit}",
+                "exit_destination": "Take exit {exit} towards {destination}"
             }
         },
         "on ramp": {
--- a/test/fixtures/v5/continue/uturn_exit.json
+++ b/test/fixtures/v5/continue/uturn_exit.json
@@ -1,0 +1,22 @@
+{
+    "step": {
+        "maneuver": {
+            "type": "continue",
+            "modifier": "uturn"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "180°-Wendung auf Way Name",
+        "en": "Make a U-turn onto Way Name",
+        "es": "Haz un cambio de sentido en Way Name",
+        "fr": "Faire demi-tour sur Way Name",
+        "id": "Putar balik ke arah Way Name",
+        "nl": "Keer om naar Way Name",
+        "ru": "Развернитесь на Way Name",
+        "sv": "Gör en U-sväng in på Way Name",
+        "vi": "Quẹo ngược lại Way Name",
+        "zh-Hans": "调头上Way Name"
+    }
+}
--- a/test/fixtures/v5/continue/uturn_exit_destination.json
+++ b/test/fixtures/v5/continue/uturn_exit_destination.json
@@ -1,0 +1,23 @@
+{
+    "step": {
+        "maneuver": {
+            "type": "continue",
+            "modifier": "uturn"
+        },
+        "name": "Way Name",
+        "destinations": "Destination 1,Destination 2",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "180°-Wendung Richtung Destination 1",
+        "en": "Make a U-turn towards Destination 1",
+        "es": "Haz un cambio de sentido hacia Destination 1",
+        "fr": "Faire demi-tour en direction de Destination 1",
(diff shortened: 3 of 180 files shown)


languages/translations/en.json
```diff
@@ -238,37 +238,51 @@
...
+                "destination": "Take the ramp towards {destination}",
+                "exit": "Take exit {exit}",
+                "exit_destination": "Take exit {exit} towards {destination}"
             },
             "left": {
                 "default": "Take the ramp on the left",
                 "name": "Take the ramp on the left onto {way_name}",
-                "destination": "Take the ramp on the left towards {destination}"
+                "destination": "Take the ramp on the left towards {destination}",
+                "exit": "Take exit {exit} on the left",
```
willwhite: I'm explicit about "on the left" here because left exits are more unusual.

languages/translations/en.json
```diff
@@ -238,37 +238,51 @@
             "default": {
                 "default": "Take the ramp",
                 "name": "Take the ramp onto {way_name}",
-                "destination": "Take the ramp towards {destination}"
+                "destination": "Take the ramp towards {destination}",
+                "exit": "Take exit {exit}",
+                "exit_destination": "Take exit {exit} towards {destination}"
```
willwhite: Wasn't sure about using an `_` here style wise. There isn't a precedent for complex/multiple fields.

test/fixtures/v5/continue/uturn_exit.json
```diff
@@ -0,0 +1,22 @@
...
+        "maneuver": {
+            "type": "continue",
+            "modifier": "uturn"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "180°-Wendung auf Way Name",
+        "en": "Make a U-turn onto Way Name",
```
daniel-j-h: Hm, thinking about situations for which a u-turn taking an exit occurs. As it's a complex maneuver for the driver do we really not want to add the exits information in addition here?

test/fixtures/v5/continue/uturn_exit_destination.json
```diff
@@ -0,0 +1,23 @@
...
+            "type": "continue",
+            "modifier": "uturn"
+        },
+        "name": "Way Name",
+        "destinations": "Destination 1,Destination 2",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "180°-Wendung Richtung Destination 1",
+        "en": "Make a U-turn towards Destination 1",
```
daniel-j-h: Same here

test/fixtures/v5/end_of_road/left_exit.json
```diff
@@ -0,0 +1,22 @@
+{
+    "step": {
+        "maneuver": {
+            "type": "end of road",
+            "modifier": "left"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
```
daniel-j-h: Not sure when end-of-road can occur in combination with an exit - these may be good candidates for validation handlers? cc @MoKob

test/fixtures/v5/fork/left_exit.json
```diff
@@ -0,0 +1,22 @@
...
+        "maneuver": {
+            "type": "fork",
+            "modifier": "left"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "Links halten an der Gabelung auf Way Name",
+        "en": "Keep left at the fork onto Way Name",
```
daniel-j-h: Is there a reason we don't add the exit information here? I can see how the "keep left" and the exit information might seem redundant but wouldn't it be a confirmation for the user to take a specific exit while keeping left?

test/fixtures/v5/merge/left_exit.json
```diff
@@ -0,0 +1,22 @@
+{
+    "step": {
+        "maneuver": {
+            "type": "merge",
+            "modifier": "left"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
```
daniel-j-h: Also a case for validation. cc @MoKob

test/fixtures/v5/notification/left_exit.json
```diff
@@ -0,0 +1,22 @@
...
+        "maneuver": {
+            "type": "notification",
+            "modifier": "left"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "Links weiterfahren auf Way Name",
+        "en": "Continue left onto Way Name",
```
daniel-j-h: Why no exit information here?

test/fixtures/v5/off_ramp/left_exit.json
```diff
@@ -0,0 +1,22 @@
...
+        "maneuver": {
+            "type": "off ramp",
+            "modifier": "left"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "Rampe auf der linken Seite nehmen auf Way Name",
+        "en": "Take exit 4A on the left",
```
daniel-j-h: Exit information is only in english translation (yet?)

test/fixtures/v5/off_ramp/uturn_exit.json
```diff
@@ -0,0 +1,22 @@
...
+    "step": {
+        "maneuver": {
+            "type": "off ramp",
+            "modifier": "uturn"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "Rampe nehmen auf Way Name",
```
daniel-j-h: German nitpick - are these really called "Rampe" and not rather "Ausfahrt"? cc @freenerd

test/fixtures/v5/rotary/default_exit.json
```diff
@@ -0,0 +1,22 @@
...
+        "maneuver": {
+            "modifier": "left",
+            "type": "rotary"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "In den Kreisverkehr fahren und auf Way Name verlassen",
+        "en": "Enter the rotary and exit onto Way Name",
```
daniel-j-h: In complex roundabouts/rotaries/traffic circles it could make sense to add the exit information if we have it.

test/fixtures/v5/fork/sharp_left_exit.json
```diff
@@ -0,0 +1,22 @@
...
+    "step": {
+        "maneuver": {
+            "type": "fork",
+            "modifier": "sharp left"
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "Scharf links abbiegen an der Gabelung auf Way Name",
```
MoKob: does this type come out of the backend? I would hope not 😨

test/fixtures/v5/merge/left_exit.json
MoKob: I don't think a `merge` and `exit` should ever be able to be in the same instruction.

test/fixtures/v5/roundabout/exit_exit.json
```diff
@@ -0,0 +1,23 @@
...
+        "maneuver": {
+            "modifier": "left",
+            "type": "roundabout",
+            "exit": 1
+        },
+        "name": "Way Name",
+        "exits": "4A,4B"
+    },
+    "instructions": {
+        "de": "In den Kreisverkehr fahren und erste Ausfahrt nehmen auf Way Name",
```
MoKob: German is kind of verbose on these. Wouldn't it be obvious to `enter the roundabout`? We could easily say `Im Kreisverkehr die erste Ausfahrt auf Way Name nehmen`, replacing `In den fahren` with `Im`.

test/fixtures/v5/fork/sharp_left_exit.json
freenerd: Do you mean that `fork + sharp left` should never occur?

test/fixtures/v5/roundabout/exit_exit.json
freenerd: Moving to https://github.com/Project-OSRM/osrm-text-instructions/issues/117

test/fixtures/v5/notification/left_exit.json
freenerd: The use case for cars here would be to have an exit on/off a ferry? Not sure if these will ever be well tagged.

test/fixtures/v5/off_ramp/uturn_exit.json
freenerd: Moving to https://github.com/Project-OSRM/osrm-text-instructions/issues/118

test/fixtures/v5/continue/uturn_exit.json
willwhite: Exits only make sense in the off-ramp case as far as I know. I would be surprised if there are any tags used in other context, but I think we should focus on the highway/ramp use case first.

languages/translations/en.json
willwhite: Makes sense, but I went in the other direction with 38c6630a3be680ad567d94dfdad50030ed68bbb0 and added specific instructions until we have enough information to simplify.

willwhite: @freenerd just saw your note further down:

> Decide on exit directions (i vote for remove left/right)

I learn toward more specificity here because it can be super jarring when there is a left exit and driver really do expect it to be specifically called out.

test/fixtures_test.js
```diff
@@ -75,6 +75,21 @@ tape.test('verify existance/update fixtures', function(assert) {
...
+        });
+        checkOrWrite(step, `${basePath}_exit`);
+
+        // exit + destination
+        step = Object.assign(clone(baseStep), {
+            name: 'Way Name',
+            destinations: 'Destination 1,Destination 2',
+            exits: '4A,4B'
+        });
+        checkOrWrite(step, `${basePath}_exit_destination`);
```
willwhite: I went with this approach because `destination` was the most obvious pattern to model this work off of. I also liked how the fixtures guarantee that exit information _isn't_ being added to instructions where it shouldn't be. I'm open to other options, but seems like the additional fixtures are easy enough to maintain.

willwhite: I also lean away from implementing this for other types and just stay focused on off ramps, which will be 99% of cases.

willwhite:
Thanks for the feedback. I'm going to merge this in so the new strings loaded into Transifex and we can get the ball rolling on translation but let me know if there are followup changes.

1ec5:
Ported to Swift in Project-OSRM/osrm-text-instructions.swift#25.

languages/translations/en.json
```diff
@@ -254,7 +254,7 @@
                 "name": "Take the ramp on the right onto {way_name}",
                 "destination": "Take the ramp on the right towards {destination}",
                 "exit": "Take exit {exit}",
-                "exit_destination": "Take exit {exit} towards {destination}"
+                "exit_destination": "Take exit {exit} on the right towards {destination}"
```
1ec5: @willwhite, if we’re saying “on the right” in the presence of a destination, why not also say “on the right” in the absence of a destination?

freenerd: I'd love to remove the direction from instructions, except when we have no way/destination/exit at all: `Take the ramp on the right`.

freenerd:
Released this as version v0.5.0 and pushed to osrm-frontend https://github.com/Project-OSRM/osrm-frontend/pull/252

Here is a route with an exit instruction http://map.project-osrm.org/?z=16&center=36.090924%2C-95.999672&loc=36.089927%2C-95.993342&loc=36.088991%2C-96.004672&hl=en&alt=0

Note that this runs against the demo server with `osrm-backend@master` and we'll need to wait for  `osrm-backend@5.9.0` until this is available in a release.

languages/translations/en.json
willwhite: "Fixed" this by removing the left/right indication in https://github.com/Project-OSRM/osrm-text-instructions/pull/124.

PR #120, opened 2017-07-06
Polish translation

malte777:
via transifex


freenerd:
Thanks @malte777

malte777:
Could you create a release? Then it would be easier to use the latest version in dependencies. Thanks, Malte

freenerd:
It's published as `v0.5.0`.

FYI: If you add a polish translations to https://github.com/Project-OSRM/osrm-frontend/tree/gh-pages/i18n we can also expose it on http://map.project-osrm.org/. Bardzo dobrze  😉

PR #124, opened 2017-07-07
Remove left/right from exit announcement.

willwhite:
Hi @freenerd. This PR does what you wanted in https://github.com/Project-OSRM/osrm-text-instructions/pull/116#discussion_r126152262.

### DIFF
--- a/languages/translations/en.json
+++ b/languages/translations/en.json
@@ -246,43 +246,43 @@
                 "default": "Take the ramp on the left",
                 "name": "Take the ramp on the left onto {way_name}",
                 "destination": "Take the ramp on the left towards {destination}",
-                "exit": "Take exit {exit} on the left",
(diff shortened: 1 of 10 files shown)


willwhite:
>  I think the danger of omitting “on the left” (or “on the right” in left-handed countries) outweighs the verbosity of “on the right” (“on the left” in left-handed countries) in the normal case.

I was leaning in this direction when I first wrote the PR but after seeing it live the verbosity is pretty offensive. The plan would be to get this straitened out once we know which countries drive on which side.

bsudekum:
Yeah, I'm in favor of keeping left/right in favor of verbosity. This information is important when the exit ramp is actually on the left, since this does actually happen.

willwhite:
Closing in favor of https://github.com/Project-OSRM/osrm-text-instructions/pull/125.

PR #125, opened 2017-07-09
Add missing 'on the right' to exit instructions.

willwhite:
Fixes consistency so we're announcing `on the right` for all instructions. The more verbose alternative to https://github.com/Project-OSRM/osrm-text-instructions/pull/124.

### DIFF
--- a/languages/translations/en.json
+++ b/languages/translations/en.json
@@ -253,7 +253,7 @@
                 "default": "Take the ramp on the right",
                 "name": "Take the ramp on the right onto {way_name}",
                 "destination": "Take the ramp on the right towards {destination}",
-                "exit": "Take exit {exit}",
+                "exit": "Take exit {exit} on the right",
                 "exit_destination": "Take exit {exit} on the right towards {destination}"
             },
             "sharp left": {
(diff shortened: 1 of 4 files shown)


freenerd:
I still think this is too verbose and we clutter 99% of cases, to cover a 1% case. My ideal design as mentioned in https://github.com/Project-OSRM/osrm-text-instructions/pull/116#discussion_r126152262:

> remove the direction (on the left/right) from instructions, except when we have no way/destination/exit at all: Take the ramp on the right.

This is because I'd prefer this project to err on the side of being as concise as possible, since an annoying, too-verbose nav to me is worse than a too-concise nav (where the user can still look at the screen if I'm unsure). I understand though that this my own opinion and there may be no objective wrong or right.

Ideally we would eventually solve this by knowing the driving-side of the current country (upstream https://github.com/Project-OSRM/osrm-backend/issues/2772) and emit the direction only on divergence of the country norm.

freenerd:
From a code side, this looks okay. Changelog entry is missing.

1ec5:
> I still think this is too verbose and we clutter 99% of cases, to cover a 1% case.

Concision is good in general, but not at the expense of omitting information potentially relevant to safe driving. I agree that left-hand exits are a minority case (in right-handed countries), but that’s precisely why it’s so important that we inform the user of a left-hand exit. Ambiguity is only part of the problem; a wrong-sided exit requires a lot of additional time to get into the correct lane, because it branches from the fastest lane rather than the slowest.

As a thought experiment, let’s suppose we gain the ability to tell whether the left-hand exit is an anomaly, but a particular instruction needs to be abridged to fit in the allotted time (mapbox/mapbox-navigation-ios#331). In that case, it might be better to omit the exit number or destination (assuming there’s a ref) than to omit the fact that the exit is on the wrong side.

PR #105, opened 2017-05-05
Added Romanian language.

mdomnita:
Added Romanian language and Romanian translation file.

Issue #136, opened 2017-08-21
Improving Swedish translation (Exit numbers, overrides, modifiers and roundabouts and questions)


davols:
I'm currently working in transifex. And having some issues and questions regarding the best way to handle the translations. 

### Overrides
~~~When it comes to turns, the preposition used makes the text incorrect when it comes to sharp/slight. Is it possible to use override/sv.js for this and delete sharp/slight for every turn? For example:~~~

`delete content.v5.turn['sharp right'];` 

~~~It seems better to use overrides to fix this (to remove `sharp` and `slight`, same for `continue` as well. Tried to have an override similar to the example above but running the `UPDATE=1 npm test` still produce turn sharp left fixtures.~~~


~~~Would I be using overrides correctly with the example above? And would it in that example just fall back to `v5.turn.right` when `sharp.right` comes up?~~~ Fixed in transifex by just translating `slight left` et al to `left` (note the uturn issue for turns)

Note for some docs: Explain the overrides a bit, preferably with examples (saw German has a several overrides)

### Exit numbering
~~~Exit numbering for highways, ignores the exit number and only uses "towards ..." Transifex is saying every line is translated. This seems to be missing? (to translate those strings)~~~ Fixed in transifex


### "extra modifiers"
~~~Related to overrides, there are several situations where one wants to remove the sharp/slight. But there are also examples where another preposition should be added between `sharp` and `left`. Never used transifex before and not sure if this is done easily.~~~ Removed the adjectives
~~~Like `turn 1{modifier:1} to 1{modifier:2}` (where `modifier1` would be sharp or left and `modifier2` right/left.~~~

Is this possible? Otherwise the "extra" modifiers (sharp/slight) should probably be removed. 

### Roundabout & Rotary
~~~Have some thoughts about the translation and also shortening it.~~~

~~~For instance the text "Enter the roundabout and take the 2nd exit towards" could be shorten. One way to translate and shorten it is "In the roundabout take the 2nd exit towards"~~~

~~~Similar for rotary which right now is "Drive into the roundabout" (in Swedish being quite literal) which could be changed to "In the roundabout"~~~ Shortened and improved in transifex

~~~Not sure about the thoughts about things like that and also where they should be changed (transifex vs override) - They would make more sense but I'm also afraid they will be incorrect for some small use cases.~~~ For Swedish they are in transifex now (for sharp and slight).

1ec5:
[...]

PR #137, opened 2017-08-22
Italian, Brazilian Portuguese, Ukrainian

1ec5:
Added Italian, Brazilian Portuguese, and Ukrainian localizations from Transifex. Also updated Russian, Spanish, ~~Swedish~~, and Vietnamese, which have complete localizations. I avoided bringing in updates to the other localizations, since there are some key untranslated strings around exits. I also omitted Romanian, which is the subject of #105.

@nicomazz @leonardortlima @Andygol @oxidase, would you mind reviewing the test fixtures to ensure that the generated strings line up with your expectations? Thank you!

/cc @bsudekum


languages/translations/ru.json
```diff
@@ -238,37 +243,51 @@
             "default": {
                 "default": "Сверните на съезд",
                 "name": "Сверните на съезд на {way_name}",
-                "destination": "Сверните на съезд в направлении {destination}"
+                "destination": "Сверните на съезд в направлении {destination}",
+                "exit": "Сверните на съезд {exit} ",
```
yuryleb: Extra trailing whitespace - I hope this is not a problem?

languages/translations/ru.json
```diff
@@ -238,37 +243,51 @@
...
+                "exit_destination": "Сверните на съезд {exit} в направлении {destination}"
             },
             "left": {
                 "default": "Сверните на левый съезд",
                 "name": "Сверните на левый съезд на {way_name}",
-                "destination": "Сверните на правый съезд в направлении {destination}"
+                "destination": "Сверните на правый съезд в направлении {destination}",
```
yuryleb: But here is "right exit" instead of "left". I should update translation :(

languages/translations/ru.json
```diff
@@ -238,37 +243,51 @@
...
+                "exit": "Сверните на съезд {exit} справа",
+                "exit_destination": "Сверните на съезд {exit} справа в направлении {destination}"
             },
             "sharp left": {
                 "default": "Поверните на левый съезд",
                 "name": "Поверните на левый съезд на {way_name}",
-                "destination": "Поверните на левый съезд в направлении {destination}"
+                "destination": "Поверните на левый съезд в направлении {destination}",
+                "exit": "Сверните на съезд {exit} слева",
```
yuryleb: Different verbs from this line and below, they also should be updated

yuryleb:
I reviewed Russian translation and updated it more once on Transifex - there were some quite critical style mistakes, sorry :disappointed:

test/fixtures/v5/continue/sharp_left_default.json
```diff
@@ -12,10 +12,13 @@
         "es": "Continua cerrada a la izquierda",
         "fr": "Continuer franchement à gauche",
         "id": "Terus tajam kiri",
+        "it": "Continua Leggera svolta a sinistra",
```
nicomazz: There are problems with the Italian translation of "sharp ...", which does not sound very good at this time.
Could I have a (street?) example where this instruction should be used?

test/fixtures/v5/end_of_road/sharp_left_default.json
```diff
@@ -12,10 +12,13 @@
         "es": "Gire a cerrada a la izquierda",
         "fr": "Tourner franchement à gauche",
         "id": "Belok tajam kiri",
+        "it": "Gira a Leggera svolta a sinistra",
```
nicomazz: This doesn't sound good. At present I do not know how to translate it better. I'll think about it.

test/fixtures/v5/fork/straight_default.json
```diff
@@ -12,10 +12,13 @@
         "es": "Mantengase recto en el cruce",
         "fr": "Rester tout droit à l'embranchement",
         "id": "Tetap lurus di pertigaan",
+        "it": "Mantieni la Continua dritto al bivio",
```
nicomazz: There is a problem with italian.
With left or right modifier "keep {modifier} at the fork" it's ok.
With "straight" modifier the phrase should be like "{modifier} at the fork" and not "Keep {modifier} at the fork" (translated in italian: "Continua dritto al bivio", without the initial "keep")

test/fixtures/v5/merge/straight_destination.json
```diff
@@ -13,10 +13,13 @@
         "es": "Gire a recto hacia Destination 1",
         "fr": "Rejoindre tout droit en direction de Destination 1",
         "id": "Bergabung lurus menuju Destination 1",
+        "it": "Immettiti Continua dritto verso Destination 1",
```
nicomazz: Same problem here with "straight" modifier

test/fixtures/v5/continue/sharp_left_default.json
1ec5: I think a sharp left would always be given as a `turn` instruction rather than a `continue` instruction, so it might be safe to ignore the weird output here. Is that correct @bsudekum?

test/fixtures/v5/end_of_road/sharp_left_default.json
1ec5: If it isn’t possible to translate the component strings in a way that would make sense, it’s also possible to override this particular configuration using an [override file](https://github.com/Project-OSRM/osrm-text-instructions/tree/master/languages/overrides/).

test/fixtures/v5/fork/straight_default.json
1ec5: This is also awkward in English: “Keep straight at the fork”. However, I think it’s unlikely that the Directions API would ever output this instruction. I think the API would omit the instruction entirely, unless the destination street has a different name, in which there would be a `continue` instruction. @bsudekum, is that correct?

languages/translations/uk.json
```diff
@@ -0,0 +1,421 @@
...
+                "name": "Рухайтесь {modifier} на {way_name}",
+                "destination": "Рухайтесь {modifier} у напрямку {destination}"
+            },
+            "uturn": {
+                "default": "Здійсніть розворот",
+                "name": "Здійсніть розворот на {way_name}",
+                "destination": "Здійсніть розворот у напрямку {destination}"
+            }
+        },
+        "off ramp": {
```
Andygol: @oxidase Please take a look here.
It confuses me such we have equal translation for `off ramp` and `on ramp` further

oxidase: @Andygol it looks like direct translation from English where no difference in `on`/`off` ramp  instructions. what confuses me is `ramp` translation as `естакада`  probably `в'їзд` / `з'їзд` is better

Andygol: @oxidase yes, I agreed with you, that `в'їзд` / `з'їзд` is much better. May you make this changes right here on github? I don't have the right to edit here.

1ec5: If any changes are needed, please make them [in Transifex](https://www.transifex.com/project-osrm/osrm-text-instructions/language/uk/), and I’ll update this PR. Otherwise, if we make the changes directly in GitHub, they’ll get overridden the next time someone pulls translations from Transifex.

test/fixtures/v5/continue/sharp_left_default.json
nicomazz: I think the terms "sharp" and "slight" could ignore in Italian, translating "sharp right" into only "right". Is this a good solution? Otherwise there would be many problems..

test/fixtures/v5/end_of_road/sharp_left_default.json
nicomazz: I've made some changes in transifex. Maybe I've resolved the problem

languages/translations/uk.json
Andygol: @1ec5 I updated Ukrainian translation on Transifex according to @oxidase suggestions

1ec5:
I completed a few remaining translations in Spanish and added it to this PR in 97c561ed490a953eaccca8c1004d3b196ac36e4b. (This PR still omits the 🇪🇸 Spanish localization that’s ongoing.) The biggest change is that the instructions use _usted_ form instead of _tuteo_. @Guardiola31337, can you take a look at the updated Spanish strings? Thanks!

1ec5:
I also completed the German localization and added it in 97c561ed490a953eaccca8c1004d3b196ac36e4b – would appreciate a quick look-over from @daniel-j-h.

leonardortlima:
I made a few more changes on transifex. I think Brazilian Portuguese is good to go now.

1ec5:
German issues fixed in 69290edd843bc4235e913b1c702226059f675743. There’s some ongoing discussion about Swedish in #136, but it sounds like this PR still represents an improvement over what’s in master.

/cc @frederoni

languages/translations/ru.json
```diff
@@ -238,37 +243,51 @@
...
+                "name": "Поверните направо на съезд на {way_name}",
+                "destination": "Поверните направо на съезд в направлении {destination}",
+                "exit": "Поверните направо на съезд {exit}",
+                "exit_destination": "Поверните направо на съезд {exit} в направлении {destination}"
             },
             "slight left": {
-                "default": "Плавно сверните на левый съезд",
-                "name": "Плавно сверните на левый съезд на {way_name}",
-                "destination": "Плавно сверните на левый съезд в направлении {destination}"
+                "default": "Перестройтесь левее съезд",
```
yuryleb: Preposition is missing

yuryleb:
Russian translation had one string with missing prepostion, Transifex was updated again.

1ec5:
I’ve removed Swedish from this PR, now that #138 has landed.

PR #138, opened 2017-08-25
Updated Swedish

davols:
# Issue

What Github issue is this PR targeting?
Regarding issue #136 - Moved to another PR in order to keep a discussion regarding the changes in here instead.  

## Misc

Removed the adjectives for turns in transifex. Default for turn works for all entries except for uturns. 
cc @frederoni @1ec5

### DIFF
--- a/languages/translations/sv.json
+++ b/languages/translations/sv.json
@@ -160,9 +160,9 @@
         },
         "merge": {
             "default": {
-                "default": "Körfältsbyte åt {modifier}",
-                "name": "Körfältsbyte åt {modifier} in på {way_name}",
-                "destination": "Körfältsbyte åt {modifier} mot {destination}"
+                "default": "Byt till {modifier} körfält",
+                "name": "Byt till {modifier} körfält in på {way_name}",
+                "destination": "Byt till {modifier} körfält mot {destination}"
             },
             "slight left": {
                 "default": "Byt till vänstra körfältet",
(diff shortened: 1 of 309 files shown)


languages/translations/sv.json
```diff
@@ -165,24 +165,24 @@
                 "destination": "Körfältsbyte åt {modifier} mot {destination}"
```
frederoni: Can we change these too or would the definite form adjective make the noun clash?
Something like {modifier.definiteForm} vänster/vänstra would be convenient.

davols: I would like to change these too. But I'm not aware if it's even possible. Since the modifier needs to be `vänster` for other instructions. Guess there is no way to extend the modifier to have a form too? But this should be changed since it's an instruction without a verb.

frederoni:
This reads so much better. Great job.
I think you have to update the test fixtures `UPDATE=1 npm test` and commit the changes.

davols:
> I think you have to update the test fixtures UPDATE=1 npm test and commit the changes.
Will do, saw the fail CI. Checked some other closed PRs for updated languages and they didn't seem to have the textures. But adding them now.

davols:
Some future improvements I guess would be:
* Have definite form for modifiers
* have a specific string for turn.uturn (see below)
`turn` with something else than strict `turn.left` & `turn.right` reads correct except for only `turn.uturn` - the only way to fix that (?) would be to change `turn.default` but that would break everything except `uturn`. So either 4 incorrect strings or 1.

1ec5:
Thank you for these improvements, @davols! We can track further improvements in #136 and Transifex.

Issue #88, opened 2016-11-20
Distance-based instructions


1ec5:
Turn-by-turn navigation applications need to say things like “In {distance}, {instruction}” (as the user approaches a maneuver) and “Continue on {name} for {distance}” (immediately after the user completes a maneuver). It’s easy enough for the application to provide these strings itself, but then it may not be localized as broadly as the strings in this project, and `{name}` won’t reflect osrm-text-instructions smarts like #77.

/cc @freenerd @bsudekum

PR #141, opened 2017-09-01
Add phrases for post instruction processing

mcwhittemore:
There are times when one wants to link two instructions or prefix them with a message. These `phrases` should be translated with the same level of detail as the rest of this project, so I think they should live here. That said, I'm not quite sure how they should be added to the translation files so currently I've only added these to `en.json`.

This PR adds a `phases` to `v5` of the translations along with these new continue phrases.

- continue.default.distance
- continue.default.namedistance
- continue.straight.distance

# Issue

I think this will close https://github.com/Project-OSRM/osrm-text-instructions/issues/88.

## Tasklist

 - [x] Add changelog entry
 - [ ] Review

### DIFF
--- a/languages/translations/en.json
+++ b/languages/translations/en.json
@@ -50,6 +50,11 @@
                 "destination": "Take the ferry towards {destination}"
             }
         },
+        "phrase": {
+            "two linked by distance": "{instruction_one} then in {distance} {instruction_two}",
+            "two linked": "{instruction_one} then {instruction_two}",
+            "one in distance": "In {distance}, {instruction_one}"
+        },
         "arrive": {
             "default": {
                 "default": "You have arrived at your {nth} destination"
@@ -81,12 +86,15 @@
                 "default": "Continue {modifier}",
                 "name": "Continue {modifier} to stay on {way_name}",
                 "destination": "Continue {modifier} towards {destination}",
-                "exit": "Continue {modifier} onto {way_name}"
+                "exit": "Continue {modifier} onto {way_name}",
+                "distance": "Continue for {distance}",
+                "namedistance": "Continue on {way_name} for {distance}"
             },
             "straight": {
                 "default": "Continue straight",
                 "name": "Continue straight to stay on {way_name}",
-                "destination": "Continue towards {destination}"
+                "destination": "Continue towards {destination}",
+                "distance": "Continue straight for {distance}"
             },
             "slight left": {
                 "default": "Continue slightly left",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Change Log
 All notable changes to this project will be documented in this file. For change log formatting, see http://keepachangelog.com/
 
+# 0.6.0
+
+- Adds `phrase` types to the English localiztion. [#141](https://github.com/Project-OSRM/osrm-text-instructions/pull/141)
+- Adds `distance`, `name` and `namedistance` options to the continue and continue straight instructions. [#141](https://github.com/Project-OSRM/osrm-text-instructions/pull/141)
+- Adds `tokenize` to the top level api so that external users can fill in osrm-text-instructions template strings. [#141](https://github.com/Project-OSRM/osrm-text-instructions/pull/141)
+
 # 0.5.4 2017-09-06
 
 - Improves the wording of "continue" instructions. [#142](https://github.com/Project-OSRM/osrm-text-instructions/pull/142)
--- a/index.js
+++ b/index.js
@@ -180,23 +180,31 @@ module.exports = function(version, _options) {
 
             // Replace tokens
             // NOOP if they don't exist
-            instruction = instruction
-                .replace('{way_name}', wayName)
-                .replace('{destination}', (step.destinations || '').split(',')[0])
-                .replace('{exit}', (step.exits || '').split(';')[0])
-                .replace('{exit_number}', this.ordinalize(language, step.maneuver.exit || 1))
-                .replace('{rotary_name}', step.rotary_name)
-                .replace('{lane_instruction}', laneInstruction)
-                .replace('{modifier}', instructions[language][version].constants.modifier[modifier])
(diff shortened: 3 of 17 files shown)


languages/translations/en.json
```diff
@@ -50,6 +50,11 @@
                 "destination": "Take the ferry towards {destination}"
             }
         },
+        "phrase": {
+            "two linked by distance": "{instruction_one} then in {distance} {instruction_two}",
```
bsudekum: I think the spaces should be removed here.

mcwhittemore: I don't see any camel casing in here, so I've merged this like this.

`two linked by distance` => `twolinkedbydistance`

bsudekum:
@mcwhittemore what is your plan for creating two instructions as in `{instruction} then {instruction}`? Right now the API only accepts a single step.

languages/translations/en.json
```diff
@@ -50,6 +50,11 @@
                 "destination": "Take the ferry towards {destination}"
             }
         },
+        "phrase": {
+            "twolinkedbydistance": "{instruction_one} then in {distance} {instruction_two}",
+            "twolinked": "{instruction_one} then {instruction_two}",
+            "oneindistance": "In {distance}, {instruction_one}"
```
miccolis: Should there be a comma here? "In 100 meters, turn right" or "In 100 meters turn right"

mcwhittemore:
> what is your plan for creating two instructions as in {instruction} then {instruction}? Right now the API only accepts a single step.

@bsudekum - I'm doing something like where `merged` is the output I'm interested in.

```
var osrmTextInstructions = require('osrm-text-instructions');
var templateString = require('osrm-text-instructions/languages/translations/en.json')['v5']['phrase']['twolinked'];
var steps = [ ... ];

var one = osrmTextInstructions.complie('v5', { language: 'en', step: steps[0] });
var two = osrmTextInstructions.complie('v5', { language: 'en', step: steps[1] });

var merged = osrmTextInstructions.tokenize(templateString, {instruction_one: one, instruction_two: two}, 'en');
```

languages/translations/en.json
mcwhittemore: I think that if you can drop the introductory clause it should be separated from the rest of the sentence by a comma. At least, that is how I remember @lyzidiamond explaining these things to me a few months back, but I'm pretty bad at grammar so I could very well be wrong here.

languages/translations/en.json
bsudekum: @mcwhittemore retracting my comment, this should spaces.

languages/translations/en.json
lyzidiamond: Exactly right @mcwhittemore! "In 100 meters, turn right" is the correct structure.

yuryleb:
BTW there are some issues with instructions combine this way:

- All phrases in text resources should start with non-capital letter (maybe except German) to be ready to insert into combined instruction into any position
- Even this doesn't help now - osrmTextInstructions.compile() will capitalize first letter by default (see meta.capitalizeFirstLetter option in language JSON files) for each instruction again
- What's about translate strings for {distance}? Meters/miles question is interesting too :wink:

mcwhittemore:
> All phrases in text resources should start with non-capital letter

Ah. Good point. I've updated my example above to use `toLowerCase()` to convert the result of `compile` into a string that will work for `tokenize`. That said, this is rather complicated and, as of now, is not something my PR tries to help the end user solve.

> What's about translate strings for {distance}

This is a good idea. I'm going to add `constants.distanceunits` which will be an object of words like `feet`, `meter` and so on.

mcwhittemore:
Hmm... it looks like all constants have to be translated before tests will pass. Is this by design?

yuryleb:
How distance value will be rounded before formatting? It will be language-dependent rule/expression (for English always in miles/feet, in kilometers/meters otherwise)?

Also I hope final distance value will be rounded to 10 :wink: Otherwise I expect "plural" hell with Russian distance units translation and also with some other languages.

mcwhittemore:
> How distance value will be rounded before formatting?

This PR is pushing this question to the down stream user and requiring them to know if they need the plural or non-plural word per the language being used.

> Also I hope final distance value will be rounded to 10 😉 Otherwise I expect "plural" hell with Russian distance units translation and also with some other languages.

Can you explain this a bit more? Are you saying you hope 1.3 meters is rounded to 10 meters?

yuryleb:
I meant there are 3 forms of quantity in Russian not only "one" and "many". For example "In xxx meters do something ..." in Russian:

- Через 21 **метр** ...
- Через 22 **метра** ...
- Через 23 **метра** ...
- Через 24 **метра** ...
- Через 25 **метров** ...
- the same up to 30 and again (but 11-20 values have the same plural quantity)

But if distance will be always rounded to 10,  the only one plural form could be used in all languages AFAIK.

Also how precise should be the distance? Again for example Garmin navigators have limited set of similar distance instructions: for 100 m, 200 m and so on till 800 m, than for 1 km, 1.5 km (BTW single word numeral in Russian) and 2 km.  I don't propose to follow exactly this way but some similar distance rounding should be used.  And if distance is less than 10 meters (or better 100 meters for drivers), it should be excluded at all.

mcwhittemore:
@yuryleb - thanks for your feedback. For now I'm going to drop unit translation support from this PR and, as a next step, start looking for a javascript lib that does this well already. My hope is that we find one this is really good. If its missing languages that we support, we can then push that common lib forward rather then writing our own.

mcwhittemore:
Oh, as a final note, the phrases currently should avoid plurality in translation as they expect the client to pass them a translated distance string.

1ec5:
Opened Project-OSRM/osrm-text-instructions.swift#37 to track a Swift port of this work.

1ec5:
Since the iOS navigation SDK has historically vended its own phrase format strings, we can copy some of the translations over in Transifex from [this resource](https://www.transifex.com/mapbox/mapbox-navigation-ios/LocalizableStrings/) to [this resource](https://www.transifex.com/project-osrm/osrm-text-instructions/enjson/).

PR #142, opened 2017-09-06
"Continue onto" --> "Continue to stay on"

### LINKED ISSUE #41: "Stay on" when continuing on the same road
Along the same lines as #30, it would be nice if we could say "Keep left to stay on Main St" if Main St. is the street both before and after the fork in the road.

/cc @freenerd @bsudekum @mayagao

lyzidiamond:
# Issue

Closes #41 -- rephrases instruction from "continue onto" to "continue to stay on" when continuing. Note that I added a special condition for exits.

## Tasklist

 - [x] update language in `en.json`
 - [x] Add changelog entry
 - [ ] Review by @willwhite and/or @bsudekum and/or @1ec5? (not sure who to tag here)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to this project will be documented in this file. For change log formatting, see http://keepachangelog.com/
 
+# 0.5.4 2017-09-06
+
+- Improves the wording of "continue" instructions. [#142](https://github.com/Project-OSRM/osrm-text-instructions/pull/142)
+
 # 0.5.3 2017-08-25
 
 - Adds Brazilian Portuguese, Italian, and Ukrainian localizations. [#137](https://github.com/Project-OSRM/osrm-text-instructions/pull/137)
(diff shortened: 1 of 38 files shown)


lyzidiamond:
@mcwhittemore Per https://github.com/Project-OSRM/osrm-text-instructions/issues/41#issuecomment-319566754 + the OSRM docs, "Continue" means "Turn in direction of modifier to stay on the same road." With that in mind, consider these scenarios:

- You are driving on Main Street. In order to stay on Main Street, you need to turn left. Prior to this PR, you would get: `"Continue left onto Main Street."` But you're **already on** Main Street. With this PR, you would get: `"Continue left to stay on Main Street."` It sounds like you're asking if we can drop "to stay", which leaves: `"Continue left on Main Street."` This sounds confusing to me -- it's ambiguous about whether you're staying on the same street or turning onto a new street.

- The one place this doesn't make sense is when taking a highway exit. Per the tests, going from a ramp onto a road is considered a `continue` maneuver, but you are actually going from one path (the ramp) to another (the road). In that case, instead of hearing `"Continue straight to stay on Main Street"` (because you're not yet on Main Street when you get to that maneuver) you will hear `"Continue straight onto Main Street"` which makes more sense.

PR #146, opened 2017-09-07
Test phrases

### LINKED ISSUE #143: Test phrases
The phrases added in #141 should be subject to our string generation tests, since they’re all format strings.

/cc @mcwhittemore

1ec5:
Added a convenience method for tokenizing a phrase, plus tests and test fixtures against that method.

Fixes #143.

/cc @mcwhittemore @lyzidiamond

### DIFF
--- a/index.js
+++ b/index.js
@@ -199,44 +199,6 @@ module.exports = function(version, _options) {
 
             return this.tokenize(language, instruction, replaceTokens);
         },
-        compilePhrase: function(language, phrase, options) {
-            if (!language) throw new Error('No language code provided');
-            if (languages.supportedCodes.indexOf(language) === -1) throw new Error('language code ' + language + ' not loaded');
-            if (!phrase) throw new Error('No phrase code provided');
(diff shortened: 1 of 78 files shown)


yuryleb:
It's not related to this PR directly but is it possible also to change arguments order of new tokenize(instruction, replaceTokens, language) function to move `language` argument first as in all other index.js's methods?

1ec5:
@yuryleb, thanks for the suggestion. I opened #149 with that change.

1ec5:
I understand your hesitation around the additional `compilePhrase()` method. I ended up adding it mainly because it was necessary for unit testing phrase generation. I’d be fine with considering it a private method and not documenting it for clients (although it still needs to be exported for the tests).

mcwhittemore:
Can't the tests just call `compile` and then `tokenize`?

1ec5:
> Can't the tests just call `compile` and then `tokenize`?

Good point. Done.

PR #147, opened 2017-09-07
Add commas to linked phrases

1ec5:
Added commas to the linked phrase strings, for a more natural cadence and consistency with the iOS navigation SDK:

https://github.com/mapbox/mapbox-navigation-ios/blob/447ccaa73d4e1d8a7578e40d86cb5e59df5f2d56/MapboxCoreNavigation/Resources/Base.lproj/Localizable.strings#L7-L11

Once #146 lands, the stub translations and test fixtures will need to be updated.

/cc @mcwhittemore @lyzidiamond

### DIFF
--- a/languages/translations/en.json
+++ b/languages/translations/en.json
@@ -51,8 +51,8 @@
             }
         },
         "phrase": {
-            "two linked by distance": "{instruction_one} then in {distance} {instruction_two}",
-            "two linked": "{instruction_one} then {instruction_two}",
+            "two linked by distance": "{instruction_one}, then in {distance}, {instruction_two}",
+            "two linked": "{instruction_one}, then {instruction_two}",
             "one in distance": "In {distance}, {instruction_one}",
             "name and ref": "{name} ({ref})"
         },
--- a/languages/translations/en.json
+++ b/languages/translations/en.json
@@ -51,7 +51,7 @@
             }
         },
         "phrase": {
-            "two linked by distance": "{instruction_one}, then in {distance}, {instruction_two}",
+            "two linked by distance": "{instruction_one}, then, in {distance}, {instruction_two}",
             "two linked": "{instruction_one}, then {instruction_two}",
             "one in distance": "In {distance}, {instruction_one}",
             "name and ref": "{name} ({ref})"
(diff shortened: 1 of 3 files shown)


languages/translations/en.json
```diff
@@ -51,8 +51,8 @@
             }
         },
         "phrase": {
-            "two linked by distance": "{instruction_one} then in {distance} {instruction_two}",
-            "two linked": "{instruction_one} then {instruction_two}",
+            "two linked by distance": "{instruction_one}, then in {distance}, {instruction_two}",
```
lyzidiamond: In my dream world, this would be: `"{instruction_one}. Then, in {distance}, {instruction_two}."` But maybe I'm being too pedantic?

languages/translations/en.json
```diff
@@ -51,8 +51,8 @@
...
+            "two linked": "{instruction_one}, then {instruction_two}",
```
lyzidiamond: Similarly, my brain wants `"{instruction_one}. Then, {instruction_two}."`

languages/translations/en.json
1ec5: I hear you. I tend to omit the comma between “then” and “in” in more casual writing but include it in more formal writing. (Likewise with pauses in speech.) In this case, I suspect that the extra comma would make the synthesized speech a bit too choppy.

1ec5:
Here’s how Amazon Polly’s Joanna voice pronounces the `two linked by distance` string:

* [Make a U-turn then in 3 feet make a U-turn](https://www.dropbox.com/s/oib6xdwv0bql3hm/nocomma.mp3?dl=0)
* [Make a U-turn, then in 3 feet, make a U-turn](https://www.dropbox.com/s/scgt3wlhkki6fks/somecomma.mp3?dl=0)
* [Make a U-turn, then, in 3 feet, make a U-turn](https://www.dropbox.com/s/scgt3wlhkki6fks/somecomma.mp3?dl=0)

languages/translations/en.json
lyzidiamond: I agree with you!

mcwhittemore:
> Make a U-turn, then, in 3 feet, make a U-turn

Sounds best.

1ec5:
dd5df5f9e833cdfd21884de22b02624787cf1b0c adds a comma between “then” and “in”.

diagram

PR #148, opened 2017-09-07
Make parentheses around ref localizable

### LINKED ISSUE #131: Parentheses around ref should be localizable
[This line](https://github.com/Project-OSRM/osrm-text-instructions/blob/c2c8b2e90964d4508d5e6bbcc60f0c7cab5401b2/index.js#L143) wraps the ref in a name with hardcoded parentheses. Chinese and Japanese use `（123）` instead of `(123)`, so we should make this format localizable.

/cc @bsudekum @freenerd

1ec5:
Added a format string for a name-ref combination, [for consistency with the iOS navigation SDK](https://github.com/mapbox/mapbox-navigation-ios/blob/447ccaa73d4e1d8a7578e40d86cb5e59df5f2d56/MapboxCoreNavigation/Resources/Base.lproj/Localizable.strings#L14).

This string is primarily meant for display. For example, Chinese uses different characters for the parentheses. However, it may affect speech synthesis in some text-to-speech engines. I’m open to moving the string to a different section of the translations file than `phrase`.

Fixes #131.

/cc @mcwhittemore @bsudekum

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file. For change 
 - Adds `phrase` types to the English localiztion. [#141](https://github.com/Project-OSRM/osrm-text-instructions/pull/141)
 - Adds `distance`, `name` and `namedistance` options to the continue and continue straight instructions. [#141](https://github.com/Project-OSRM/osrm-text-instructions/pull/141)
 - Adds `tokenize` to the top level api so that external users can fill in osrm-text-instructions template strings. [#141](https://github.com/Project-OSRM/osrm-text-instructions/pull/141)
+- Made the format of name-and-ref combinations localizable. [#148](https://github.com/Project-OSRM/osrm-text-instructions/pull/141)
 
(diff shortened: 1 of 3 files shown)


yuryleb:
I'm some late but why `{name}` var here instead of `{way_name}` used in all other places? What description will you put into Transifex dictionary for {name} -"the same as {way_name} but different especially for the only string"?

yuryleb:
Moreover, I don't think this good idea at all :disappointed:  To be general solution, all `{way_name}`-containing phrases should be extended with `...{way_name} ({ref})...` variants to give translators the chance for proper translation of whole phrase, say:
```
        "continue": {
            "default": {
                "default": "Turn {modifier}",
                "name": "Turn {modifier} to stay on {way_name}",
                 "name and ref": "Turn {modifier} to stay on {way_name} ({ref})",
                "destination": "Turn {modifier} towards {destination}",
                "exit": "Turn {modifier} onto {way_name}"
            },
```

PR #150, opened 2017-09-11
Add exit rotary/roundabout

bsudekum:
Adds handling for new types `exit rotary` and `exit roundabout`. 

This can be handled as a simple turn, no special instruction needed.

/cc @1ec5 @mcwhittemore

PR #151, opened 2017-09-11
Add case for exit rotary/roundabout

bsudekum:
This fixes https://github.com/Project-OSRM/osrm-text-instructions/pull/150#issuecomment-328629147 and also adds actual translations for `exit rotary` and `exit roundabout`. This way, we do not getting logging that the type is missing. Note, these are coped from the `turn` type for each language, this is the behavior we had before in https://github.com/Project-OSRM/osrm-text-instructions/pull/150.

/cc @1ec5 @mcwhittemore


index.js
```diff
@@ -150,10 +150,10 @@ module.exports = function(version, _options) {
             if (name && ref && name !== ref && !wayMotorway) {
                 var phrase = instructions[language][version].phrase['name and ref'] ||
                     instructions.en[version].phrase['name and ref'];
-                wayName = this.tokenize(phrase, {
+                wayName = this.tokenize(language, phrase, {
```
bsudekum: Argument order got mixed up in https://github.com/Project-OSRM/osrm-text-instructions/pull/148. Strange tests passed then.

1ec5: Sorry, I missed this when committing 5f4f0734b8ff92b34263d05eae55e251698b909d at the same time for #152. Probably best to keep this PR focused on roundabouts.

bsudekum:
@1ec5 logging will when running tests or running on the server and the type is either `exit rotary` or `exit roundabout`, every time.

bsudekum:
@1ec5 updated and removed overlapping https://github.com/Project-OSRM/osrm-text-instructions/pull/152 fix.

Issue #94, opened 2017-02-22
Transifex script should use tx


1ec5:
transifex.js should use the [`tx` command line client](https://docs.transifex.com/client/), which provides a better [authentication](https://github.com/Project-OSRM/osrm-text-instructions/blob/c5bf78223c6c3c46c2886545422ed04573c91fb3/scripts/transifex.js#L9-L17) mechanism and options for some of the things we’re doing manually in that script, like mapping language codes. For inspiration, the mapbox-gl-native project uses `tx` with a [simple configuration file](https://github.com/mapbox/mapbox-gl-native/blob/master/.tx/config).

/cc @freenerd @frederoni

PR #154, opened 2017-09-13
update translations from transifex

mcwhittemore:
This is my first translation update. I think I did this right, but some :eyes: would be nice.


languages/translations/de.json
```diff
@@ -84,23 +84,34 @@
         },
         "continue": {
             "default": {
-                "default": "{modifier} weiterfahren",
-                "name": "{modifier} weiterfahren auf {way_name}",
-                "destination": "{modifier} weiterfahren Richtung {destination}"
+                "default": "Turn {modifier}",
+                "name": "Turn {modifier} to stay on {way_name}",
+                "destination": "Turn {modifier} towards {destination}",
```
1ec5: Before we merge this, let’s go into Transifex and accept any `100% match` suggestions so that perfectly acceptable translations don’t get overridden by the original English versions.

mcwhittemore: @1ec5 - I'm not sure how to do that

languages/translations/de.json
```diff
@@ -84,23 +84,34 @@
         },
         "continue": {
             "default": {
-                "default": "{modifier} weiterfahren",
-                "name": "{modifier} weiterfahren auf {way_name}",
-                "destination": "{modifier} weiterfahren Richtung {destination}"
+                "default": "{modifier} abbiegen",
```
daniel-j-h: Does "abbiegen" make sense here? It basically means "to turn", but you can get a `continue straight` which we would then translate into "Geradeaus abbiegen" -> "turn straight" (?)

languages/translations/de.json
```diff
@@ -84,23 +84,34 @@
...
+                "name": "Turn {modifier} to stay on {way_name}",
```
daniel-j-h: This is English still

languages/translations/de.json
```diff
@@ -84,23 +84,34 @@
...
+                "destination": "{modifier} abbiegen Richtung {destination}",
+                "exit": "{modifier} abbiegen auf {way_name}"
+            },
+            "sharp left": {
+                "default": "Scharf links",
+                "name": "Make a sharp left to stay on {way_name}",
```
daniel-j-h: Still English

languages/translations/de.json
```diff
@@ -84,23 +84,34 @@
...
+                "destination": "Scharf links Richtung {destination}"
```
daniel-j-h: Here and below: this feels a bit too short, it's basically

"Sharp right towards"

languages/translations/de.json
daniel-j-h: Here and below: for all the `continue` maneuvers I think we should highlight that the road actually continues.

languages/translations/de.json
```diff
@@ -84,23 +84,34 @@
...
+                "default": "Scharf rechts",
+                "name": "Make a sharp right to stay on {way_name}",
+                "destination": "Scharf rechts Richtung {destination}"
             },
             "slight left": {
-                "default": "Leicht links weiter",
-                "name": "Leicht links weiter auf {way_name}",
-                "destination": "Leicht links weiter Richtung {destination}"
+                "default": "Scharf links",
+                "name": "Make a slight left to stay on {way_name}",
```
daniel-j-h: Here and below: English still in here

languages/translations/sv.json
```diff
@@ -84,28 +84,41 @@
...
+            "sharp right": {
+                "default": "Gör en skarp högersväng",
+                "name": "Make a sharp right to stay on {way_name}",
+                "destination": "Gör en skarp högersväng mot {destination}"
             },
             "slight left": {
-                "default": "Fortsätt åt vänster",
-                "name": "Fortsätt åt vänster in på {way_name}",
-                "destination": "Fortsätt åt vänster mot {destination}"
+                "default": "Make a slight left",
```
davols: Same here, english. several of the continue has a history of having a translation previously. But now they are empty and untranslated.

languages/translations/de.json
mcwhittemore: Sounds like we need a `continue straight` override like we've done here: https://github.com/Project-OSRM/osrm-text-instructions/blob/master/languages/translations/en.json#L93

davols:
Swedish has been updated. Both the new not previously used strings are translated and also the ones that disappeared for some reason in Transifex.

Would assume that fixes the comment I made. 
Latest commit was as of this writing 7min ago - And I just finished fixing the last Swedish bits. 

Please check if there are any changes to have Swedish in this PR, or I could make my own PR for Swedish specifically. 

ping @frederoni for some extra eyes

mcwhittemore:
@davols - Thanks for the heads up.

I've pulled from Transifex again and then ran [a script](https://gist.github.com/mcwhittemore/4c64678c11313bbd21779186af19717e) to compare the post Transifex version of the translations with the English translation. Where the non-english and english matched up, I reset the value to what is currently in master. I think this should mean that the leaking of English has been fixed up. I'm scrolling through the changes now to confirm.

mcwhittemore:
> It looks like some German translation was ripped out for its English counterpart? Care to elaborate what's going on here?

I'm not 100% sure why this has happened. @1ec5 mentioned that this is partial because we blindly pull files down from Transifex and that if we [used the tx cli tool](https://github.com/Project-OSRM/osrm-text-instructions/issues/94) this would be less likely to happen.

Thanks for your eyes @davols and @daniel-j-h. If I can get one last review, I think we can merge this.

languages/translations/de.json
```diff
@@ -84,12 +84,12 @@
         },
         "continue": {
             "default": {
-                "default": "{modifier} weiterfahren",
+                "default": "{modifier} abbiegen",
                 "name": "{modifier} weiterfahren auf {way_name}",
-                "destination": "{modifier} weiterfahren Richtung {destination}"
+                "destination": "{modifier} abbiegen Richtung {destination}"
```
daniel-j-h: Why did we change from "weiterfahren" (continue) here to "abbiegen" (turn) for the `continue` instruction?

languages/translations/de.json
```diff
@@ -84,12 +84,12 @@
...
             },
             "slight left": {
-                "default": "Leicht links weiter",
+                "default": "Scharf links",
```
daniel-j-h: slight left should be "Leicht links" - Scharf means sharp, maybe a mixup in the translations

languages/translations/de.json
mcwhittemore: So, most of the `continue` maneuvers require you to make a turn so we changed the `default` to use `turn` and then created an override for `straight` to say continue. I'm guess that the translator say this change in english and made it here, but didn't provide straight override.

test/fixtures/v5/continue/left_exit.json
```diff
@@ -13,12 +13,12 @@
...
         "fr": "Continuer à gauche sur Way Name",
         "id": "Terus kiri ke Way Name",
-        "it": "Continua a sinistra in Way Name",
+        "it": "Gira a sinistra in Way Name",
         "nl": "Ga links naar Way Name",
         "pl": "Kontynuuj lewo na Way Name",
         "pt-BR": "Continue à esquerda em Way Name",
         "ru": "Двигайтесь налево по Way Name",
-        "sv": "Fortsätt vänster in på Way Name",
+        "sv": "Sväng vänsteroch fortsätt på Way Name",
```
frederoni: Nit: missing space (fixed in Transifex) /cc @davols

1ec5:
Regarding the English strings creeping in, this is unfortunately quite normal for translation systems. Whenever a base string (in English for this project) changes, all the translations of that string are “dirtied” in Transifex, so translators need to go in and update the translations. For minor changes like “Take a sharp left” → “Make a sharp left”, this should just be a matter of accepting the closest-matching suggestion. Meanwhile, any updates to this project will cause an English string to be substituted for the dirtied string, because outdated strings can be problematic (consider a change that adds or removes a placeholder token).

Replacing our ad-hoc tool with `tx` (#94) will help somewhat, as it would give us options for skipping a resource that’s inadequately translated according to Transifex. Choosing a different message format (#19) would help by giving us the ability to break up en.json into multiple resources, minimizing the disruption caused by changes to one resource.

PR #156, opened 2017-09-15
Russian translation update

yuryleb:
Small updates and fixes in new exit roundabout/rotary messages synchronized from Trasifex

### DIFF
--- a/languages/translations/ru.json
+++ b/languages/translations/ru.json
@@ -85,9 +85,9 @@
         "continue": {
             "default": {
                 "default": "Двигайтесь {modifier}",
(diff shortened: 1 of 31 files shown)


yuryleb:
It's strange, I did `UPDATE=1 npm test` :disappointed: Will try to repeat

yuryleb:
`UPDATE=1 npm test` wrote all fixtures with forced LF line breaks so I didn't find initially required changes - all fixtures were changed. Now I re-checked all thousands fixtures one by one

1ec5:
🙇 Thank you!

PR #157, opened 2017-09-16
Force LF line breaks for JS & JSON files

yuryleb:
# Issue

Current `ESLint` [settings](https://github.com/Project-OSRM/osrm-text-instructions/blob/master/.eslintrc.json#L76-L79) expect only LF line breaks in JS files; `JSON.stringify()` used by `npm run transifex` and `UPDATE=1 npm test` writes LF line breaks for generated JSON files.

This is the problem for Windows users with default `auto` line breaks option - all files after regeneration became invalid for `ESLint` and changed for `git`.

This PR adds repository option to always checkout JS and JSON files with expected LF line breaks. This is a bit inconvenient to work with them in Windows but makes further life much more easier.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.js text eol=lf
+*.json text eol=lf
--- a/scripts/transifex.js
+++ b/scripts/transifex.js
@@ -39,6 +39,6 @@ languages.supportedCodes.forEach((code) => {
         }
 
         // Write language file
-        fs.writeFileSync(`${__dirname}/../languages/translations/${code}.json`, JSON.stringify(content, null, 4));
+        fs.writeFileSync(`${__dirname}/../languages/translations/${code}.json`, JSON.stringify(content, null, 4) + '\n');
     });
 });


mcwhittemore:
Can we explain eslint to cover json too so that we don't need the `.gitattributes`?

yuryleb:
Yes, we can remove mentioned [`eslint` option](https://github.com/Project-OSRM/osrm-text-instructions/blob/master/.eslintrc.json#L76-L79) for example to don't check line breaks at all. But we can't easily change JSON.stringify() behavior to use OS-dependent line breaks and force `git` to treat text files with different line breaks unchanged. Also we can't control global `git` line breaks option - it could be `lf`, `crlf` or `auto` (by default).

So we should either carefully post-process `JSON.stringify()` output (say with `replace('\n', os.EOL)` and `require('os')` as I hope) in at least 2 places (`trasifex.js` and `index_test.js`) and suppose that Windows users always use `auto` option or just force `lf` line breaks everywhere. ~10 code lines change + 1 presumption vs 2 changed lines in this commit - what's better?

mcwhittemore:
Yury,

Sorry, my knowledge of Windows is pretty lacking, so I want to make sure I understand the problem before moving with a solution.

Right now `.eslintrc` requires that all files checked with `eslint` have unix styled lines breaks at the end of the file. The projects [lint script](https://github.com/Project-OSRM/osrm-text-instructions/blob/61a9acdea6a61fb35741cdf42e76f3e7418d6bd4/package.json#L23) only targets `.js` files which means that `.json` files can have their end of file line breaks changed without eslint throwing an error.

When a user checkouts this project on Windows with the default auto line breaks option, all files in the repo are changed to have a non-unix end of file line break. This results in eslint breaking.

Is that correct so far?

If so, I'm a bit confused about how `.json` files are creating a problem for you. From what I see, eslint doesn't evaluate this files in anywhere and thus git changing the files at checkout shouldn't result in linting errors.

If you can, please feel free to post screen shots of what you are seeing, I think that will help me understand a bit more.

yuryleb:
Yes, by default Windows checks out with CRLF and checks in with LF. No problem for repository but issue for `eslint`. But this is not so serious as:

When I do `npm run transifex` I got all language JSON files changed even if there were no real changes:
npm-run-transifex
Ok, I'm big boy, I can re-check all them one by one and discard really unchanged files manually.

But if I want to update fixture JSON files with `UPDATE=1 npm test` (actually with 2 commands in Wndows: `set UPDATE=1` and `npm test`) I got more problems.
First `npm test` call returns >300 errors like:
>   20:1    error  Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style

Ok, removing that linebreaks rule, trying again. Validation passed, all fixtures regenerated. But `git` shows all them changed:
npm-test
And checking all them one by one is much harder now, really. I missed most of really changed fixtures when I recently updated Russian strings :disappointed:

So I propose to fix this once and for all. Yes, JS and JSON can't be edited with Windows Notepad after this, but all other more-less advanced editors (like Visual Studio Code) will handle LF files with no any problems.

And no changes for *nix users and Mac's :wink:

mcwhittemore:
@yuryleb - thanks for the walk through and explaining your thoughts. I haven't worked on a project with `.gitattributes` checked in before but I'm down for seeing how this helps bring in people on other dev environments.

PR #159, opened 2017-09-18
Replace ES5 lambda in Array.some() with Array.indexOf()

yuryleb:
# Issue

Used ES5 lambda function with Array.some() introduced in #129 is incompatible with ES4-based Uglify/Browserify called by OSRM frontend.

This PR replaces Array.some() call with Array.indexOf().


yuryleb:
@bsudekum, please review proposed change - is it the same you supposed initially?

PR #162, opened 2017-09-23
Add Esperanto localization

1ec5:
This change pulls in the 100% complete Esperanto localization from Transifex.

@Verdulo, can you take a look at the test fixtures to verify that the generated strings match your expectations? Dankon!


verdulo:
I corrected some translations on transifex, please regenerate the phrases.

PR #161, opened 2017-09-23
Update Spanish, Swedish translations

1ec5:
Updated Spanish and Swedish translations from Transifex.

/cc @Sealos @davols @frederoni

PR #163, opened 2017-09-23
Add Castilian Spanish localization

1ec5:
Pulled in the 100% complete Castilian Spanish localization from Transifex.

@Sealos, can you check the test fixtures to make sure the generated strings match your expectations? You might also want to compare the Castilian Spanish translations to the generic Spanish translations, which would primarily be shown to Latin American Spanish speakers. ¡Muchas gracias!

/ref #161
/cc @Guardiola31337

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file. For change 
 
 ## master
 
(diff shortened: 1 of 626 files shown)


Sealos:
Ok, checked the generated translations and fixed some extra spaces.

1ec5:
Thanks, I pulled down the latest translations, so this is good to go.

PR #167, opened 2017-10-03
expose getWayName from compile for other use cases

mcwhittemore:
New functionality to allow users of this module to have the same `wayName` logic as used in compile.

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,6 +69,45 @@ module.exports = function(version, _options) {
 
             return config.join('');
         },
+        getWayName: function(language, step, options) {
+            var classes = options ? options.classes || [] : [];
+            if (typeof step !== 'object') throw new Error('step must be an Object');
+            if (!Array.isArray(classes)) throw new Error('classes must be an Array or undefined');
+
+            var wayName;
+            var name = step.name || '';
+            var ref = (step.ref || '').split(';')[0];
+
+            // Remove hacks from Mapbox Directions mixing ref into name
+            if (name === step.ref) {
+                // if both are the same we assume that there used to be an empty name, with the ref being filled in for it
+                // we only need to retain the ref then
+                name = '';
+            }
+            name = name.replace(' (' + step.ref + ')', '');
+
+            // In attempt to avoid using the highway name of a way,
+            // check and see if the step has a class which should signal
+            // the ref should be used instead of the name.
+            var wayMotorway = classes.indexOf('motorway') !== -1;
+
+            if (name && ref && name !== ref && !wayMotorway) {
+                var phrase = instructions[language][version].phrase['name and ref'] ||
+                    instructions.en[version].phrase['name and ref'];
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -68,6 +68,45 @@ module.exports = function(version, _options) {
...
+            var name = step.name || '';
+            var ref = (step.ref || '').split(';')[0];
+
+            // Remove hacks from Mapbox Directions mixing ref into name
+            if (name === step.ref) {
+                // if both are the same we assume that there used to be an empty name, with the ref being filled in for it
+                // we only need to retain the ref then
+                name = '';
+            }
+            name = name.replace(' (' + step.ref + ')', '');
```
yuryleb: Maybe it's better to add also `else` here - why it's necessary to search/replace in empty string?

mcwhittemore: If your thinking about perf here @yuryleb, searching an empty string should be really fast. Maybe almost as fast as the if statement check.

I try to have fewer if statements than more as a coding style (though I just copied this code from a few lines down). For this case in particular, I think avoiding the cost of String.search via an if statement is programing from the interpreter rather than other's reading the code.

index.js
```diff
@@ -127,40 +166,7 @@ module.exports = function(version, _options) {
...
-                    ref: ref
-                });
-            } else if (name && ref && wayMotorway && (/\d/).test(ref)) {
-                wayName = ref;
-            } else if (!name && ref) {
-                wayName = ref;
-            } else {
-                wayName = name;
-            }
+            var wayName = this.getWayName(language, step, options);
```
1ec5: Note that, at this point, the `grammarize()` step being added in #102 won’t have run yet. After #102 lands, the return value won’t necessarily be in the final grammatical form, and if you try to pass this value into `tokenize()`, you may end up with unexpected results. (For example, the name may be undeclined due to interference from SSML tags, or the grammar rules may be applied twice.)

mcwhittemore: @1ec5 - isn't the below code what #102 does making this pretty safe to do? As long as grammarize is called in tokenize using this to make your own `compile` should be OK from my reading.

```js
var wayName = osrmti.getWayName('ru', step);
var instructions = osrmti.tokenize('ru', 'Продолжите движение по {way_name}', {
  way_name: wayName
});
```

1ec5: This example is fine, but imagine that we manipulate `wayName` before passing it into `tokenize()`:

```js
var wayName = osrmti.getWayName('ru', step);
wayName = wayName.toUpperCase();
var instructions = osrmti.tokenize('ru', 'Продолжите движение по {way_name}', {
  way_name: wayName
});
```

Then `grammarize()` will have no effect when run as part of `tokenize()`. It’s fine for `getWayName()` to have this caveat, but I think we should make it clear via documentation that the return value should be passed into `tokenize()` without any modifications.

/ref https://github.com/Project-OSRM/osrm-text-instructions/pull/102#discussion_r142546313

PR #168, opened 2017-10-03
Add fallback code for unsupported locales

bsudekum:
This PR adds support for trying to find locales which we currently do not support. This gist of it is, that if a developer provides `es-MX`, we should give them instructions provides by `es`.

This also moves to use `en` when a language is not found.

/cc @mcwhittemore @1ec5

### DIFF
--- a/index.js
+++ b/index.js
@@ -286,8 +286,8 @@ module.exports = function(version, _options) {
             } else if (languages.instructions[languageCode]) {
                 return languageCode;
 
-            // Same language code and any script code (lng-Scpx)
-            } else if (supportedLanguageCodes.indexOf(languageCode) > -1 && scriptCode) {
+            // Same language code and any script code (lng-Scpx) and the found language contains a script
+            } else if (supportedLanguageCodes.indexOf(languageCode) > -1 && scriptCode && languages.supportedCodes[supportedLanguageCodes.indexOf(languageCode)].split('-')[1].length === 4) {
                 return languages.supportedCodes[supportedLanguageCodes.indexOf(languageCode)];
 
             // Same language code and any country code (lng-CX)
--- a/index.js
+++ b/index.js
@@ -277,6 +277,10 @@ module.exports = function(version, _options) {
                 countryCode = splitLocale[2];
             }
 
+            var supportedLanguageCode = languages.supportedCodes.map(function(locale) {
+                return locale.toLowerCase().split('-')[0];
+            });
+
             // Same language code and script code (lng-Scpt)
             if (languages.instructions[languageCode + '-' + scriptCode]) {
                 return languageCode + '-' + scriptCode;
@@ -286,16 +290,16 @@ module.exports = function(version, _options) {
                 return languageCode + '-' + countryCode;
 
             // Same language code (lng)
-            } else if (languages.instructions[languageCode]) {
-                return languageCode;
+            } else if (supportedLanguageCode.indexOf(languageCode) > -1) {
+                return languages.supportedCodes[supportedLanguageCode.indexOf(languageCode)];
 
             // Same language code and any script code (lng-Scpx)
-            } else if (languages.instructions[languageCode] && scriptCode) {
-                return languageCode;
+            } else if (supportedLanguageCode.indexOf(languageCode) > -1 && scriptCode) {
+                return languages.supportedCodes[supportedLanguageCode.indexOf(languageCode)];
 
             // Same language code and any country code (lng-CX)
-            } else if (languages.instructions[languageCode] && countryCode) {
-                return languageCode;
+            } else if (supportedLanguageCode.indexOf(languageCode) > -1 && countryCode) {
+                return languages.supportedCodes[supportedLanguageCode.indexOf(languageCode)];
             } else {
                 return 'en';
             }
--- a/index.js
+++ b/index.js
@@ -13,20 +13,17 @@ module.exports = function(version, _options) {
 
     return {
         capitalizeFirstLetter: function(language, string) {
-            return string.charAt(0).toLocaleUpperCase(this.getBestMatchingLanguage(language)) + string.slice(1);
+            return string.charAt(0).toLocaleUpperCase(language) + string.slice(1);
         },
-        ordinalize: function(originalLanguage, number) {
+        ordinalize: function(language, number) {
             // Transform numbers to their translated ordinalized value
-            if (!originalLanguage) throw new Error('No language code provided');
-
-            var language = this.getBestMatchingLanguage(originalLanguage);
+            if (!language) throw new Error('No language code provided');
 
             return instructions[language][version].constants.ordinalize[number.toString()] || '';
         },
-        directionFromDegree: function(originalLanguage, degree) {
+        directionFromDegree: function(language, degree) {
             // Transform degrees to their translated compass direction
-            if (!originalLanguage) throw new Error('No language code provided');
-            var language = this.getBestMatchingLanguage(originalLanguage);
+            if (!language) throw new Error('No language code provided');
             if (!degree && degree !== 0) {
                 // step had no bearing_after degree, ignoring
                 return '';
@@ -75,6 +72,7 @@ module.exports = function(version, _options) {
         getWayName: function(language, step, options) {
             var classes = options ? options.classes || [] : [];
             if (typeof step !== 'object') throw new Error('step must be an Object');
+            if (!language) throw new Error('No language code provided');
             if (!Array.isArray(classes)) throw new Error('classes must be an Array or undefined');
 
             var wayName;
@@ -111,10 +109,9 @@ module.exports = function(version, _options) {
 
             return wayName;
         },
-        compile: function(originalLanguage, step, options) {
-            if (!originalLanguage) throw new Error('No language code provided');
-            var language = this.getBestMatchingLanguage(originalLanguage);
-
+        compile: function(language, step, options) {
+            if (!language) throw new Error('No language code provided');
+            if (languages.supportedCodes.indexOf(language) === -1) throw new Error('language code ' + language + ' not loaded');
             if (!step.maneuver) throw new Error('No step maneuver provided');
 
             var type = step.maneuver.type;
@@ -209,10 +206,8 @@ module.exports = function(version, _options) {
 
             return this.tokenize(language, instruction, replaceTokens);
         },
-        grammarize: function(originalLanguage, name, grammar) {
-            if (!originalLanguage) throw new Error('No language code provided');
-
-            var language = this.getBestMatchingLanguage(originalLanguage);
+        grammarize: function(language, name, grammar) {
+            if (!language) throw new Error('No language code provided');
             // Process way/rotary name with applying grammar rules if any
             if (name && grammar && grammars && grammars[language] && grammars[language][version]) {
                 var rules = grammars[language][version][grammar];
@@ -231,10 +226,8 @@ module.exports = function(version, _options) {
 
             return name;
         },
-        tokenize: function(originalLanguage, instruction, tokens) {
-            if (!originalLanguage) throw new Error('No language code provided');
-
-            var language = this.getBestMatchingLanguage(originalLanguage);
+        tokenize: function(language, instruction, tokens) {
+            if (!language) throw new Error('No language code provided');
             // Keep this function context to use in inline function below (no arrow functions in ES4)
             var that = this;
             var output = instruction.replace(/\{(\w+):?(\w+)?\}/g, function(token, tag, grammar) {
--- a/index.js
+++ b/index.js
@@ -270,7 +270,7 @@ module.exports = function(version, _options) {
                 return languageCode;
             // Same language code and any script code (lng-Scpx) and the found language contains a script
             } else if (languages.parsedSupportedCodes.find(function (language) {
-                return language.language === languageCode && language.scriptCode;
+                return language.languageCode === languageCode && language.scriptCode;
             })) {
                 return languages.parsedSupportedCodes.find(function (language) {
                     return language.languageCode === languageCode && language.scriptCode;
--- a/index.js
+++ b/index.js
@@ -13,17 +13,20 @@ module.exports = function(version, _options) {
 
     return {
         capitalizeFirstLetter: function(language, string) {
-            return string.charAt(0).toLocaleUpperCase(language) + string.slice(1);
+            return string.charAt(0).toLocaleUpperCase(this.getBestMatchingLanguage(language)) + string.slice(1);
         },
-        ordinalize: function(language, number) {
+        ordinalize: function(originalLanguage, number) {
             // Transform numbers to their translated ordinalized value
-            if (!language) throw new Error('No language code provided');
+            if (!originalLanguage) throw new Error('No language code provided');
+
+            var language = this.getBestMatchingLanguage(originalLanguage);
 
             return instructions[language][version].constants.ordinalize[number.toString()] || '';
         },
-        directionFromDegree: function(language, degree) {
+        directionFromDegree: function(originalLanguage, degree) {
             // Transform degrees to their translated compass direction
-            if (!language) throw new Error('No language code provided');
+            if (!originalLanguage) throw new Error('No language code provided');
+            var language = this.getBestMatchingLanguage(originalLanguage);
             if (!degree && degree !== 0) {
                 // step had no bearing_after degree, ignoring
                 return '';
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -210,6 +211,30 @@ module.exports = function(version, _options) {
             }
 
             return output;
+        },
+        findAppropriateLanguage: function(language) {
```
1ec5: Does this method need to be exported? Also, how about calling it `getAvailableLanguage()` or `getBestMatchingLanguage()`?

index.js
```diff
@@ -68,9 +68,10 @@ module.exports = function(version, _options) {
 
             return config.join('');
         },
-        compile: function(language, step, options) {
-            if (!language) throw new Error('No language code provided');
-            if (languages.supportedCodes.indexOf(language) === -1) throw new Error('language code ' + language + ' not loaded');
+        compile: function(originalLanguage, step, options) {
+            if (!originalLanguage) throw new Error('No language code provided');
+            var language = this.findAppropriateLanguage(originalLanguage);
```
1ec5: Several other exported methods take a language parameter as well. We should be consistent; otherwise, a client using the `tokenize()` method, for example, will end up with a mismash.

index.js
```diff
@@ -210,6 +211,30 @@ module.exports = function(version, _options) {
             }
 
             return output;
+        },
+        findAppropriateLanguage: function(language) {
+            if (languages.supportedCodes.indexOf(language) > -1) {
```
1ec5: `languages.instructions` is an object with language codes as the keys. I think it’d be a bit easier to work with that object than the supported codes array.

bsudekum:
> Doing this would let us do something like _.merge({}, englishTranslation, foundTranslation) and remove all the if not in the lang, look in english checks.

index.js
```diff
@@ -210,6 +211,30 @@ module.exports = function(version, _options) {
...
+            });
+
+            if (filteredSupportedLanguages.length === 1) {
+                return filteredSupportedLanguages[0];
+            } else if (filteredSupportedLanguages.length > 1) {
+                // If more than one language is found, use the most generic version,
+                // no country code needed.
+                return filteredSupportedLanguages.sort(function(a, b) {
+                    return a.length - b.length;
+                })[0];
```
1ec5: I think we should look for matches in this order ([prior art](http://cldr.unicode.org/development/development-process/design-proposals/transform-fallback#TOC-Language-tag-fallback)), at least until we can pull in a more sophisticated library that specializes in language fallback logic:

1. Full locale code (e.g., `lng-Scpt-CC`)
1. Language code and script code (`lng-Scpt`)
1. Language code and country code (`lng-CC`)
1. Language code (`lng`)
1. Language code and any script code (`lng-Scpx`)
1. Language code and any country code (`lng-CX`)
1. English (`en`)

It’ll help if we use `languages.instructions` instead of `languages.supportedCodes`.

/cc @jcsg

index.js
bsudekum: Let's keep the list of languages we're using here consistent. An array of locales is little more accurate than a dictionary with keys IMO.

test/index_test.js
```diff
@@ -139,13 +139,72 @@ tape.test('v5 compile', function(t) {
...
+    t.test('es-MX fallback to es', function(assert) {
+        var v5Compiler = compiler('v5');
+
+        assert.equal(v5Compiler.compile('es-MX', {
+            maneuver: {
+                type: 'turn',
+                modifier: 'left'
+            },
+            name: 'Way Name'
+        }), 'Gire a la izquierda en Way Name');
```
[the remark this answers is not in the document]
1ec5: Actually, maybe it’d be better to unit test `getAvailableLanguage()` directly instead of (or in addition to) testing its integration as part of `compile()`.

index.js
jcsg: I think that sounds good. I was originally worried that condition 5 might mean that we'd serve Serbian in Latin rather than Cyrillic, but that probably would be better than just serving English for that region, especially if the voice instructions still use Serbian.

Thanks for looping me in, @1ec5 .

1ec5:
> Can we change findAppropriateLanguage to findTranslation and have this return the translation object rather than a string which is a key in that index?

We could simplify this PR considerably by removing all the calls to `getBestMatchingLanguage()` and making the client responsible for calling `getBestMatchingLanguage()` before passing a locale code into any other method like `compile()` or `tokenize()`. To enforce that, this library would throw an error if the passed-in locale code isn’t available.

index.js
```diff
@@ -243,6 +253,56 @@ module.exports = function(version, _options) {
...
+            if (languages.instructions[languageCode + '-' + scriptCode]) {
+                return languageCode + '-' + scriptCode;
+
+            // Same language code and country code (lng-CC)
+            } else if (languages.instructions[languageCode + '-' + countryCode]) {
+                return languageCode + '-' + countryCode;
+
+            // Same language code (lng)
+            } else if (supportedLanguageCode.indexOf(languageCode) > -1) {
+                return languages.supportedCodes[supportedLanguageCode.indexOf(languageCode)];
```
1ec5: This entire expression is equivalent to just `languageCode`.

index.js
```diff
@@ -243,6 +253,56 @@ module.exports = function(version, _options) {
...
+            // Same language code and script code (lng-Scpt)
+            if (languages.instructions[languageCode + '-' + scriptCode]) {
+                return languageCode + '-' + scriptCode;
+
+            // Same language code and country code (lng-CC)
+            } else if (languages.instructions[languageCode + '-' + countryCode]) {
+                return languageCode + '-' + countryCode;
+
+            // Same language code (lng)
+            } else if (supportedLanguageCode.indexOf(languageCode) > -1) {
```
1ec5: This is guaranteed to be equivalent to `if (language.instructions[languageCode])`.

index.js
```diff
@@ -243,6 +253,56 @@ module.exports = function(version, _options) {
...
+            } else if (splitLocale.length === 2 && splitLocale[1].length === 2) {
+                languageCode = splitLocale[0];
+                countryCode = splitLocale[1];
+            } else if (splitLocale.length === 3) {
+                languageCode = splitLocale[0];
+                scriptCode = splitLocale[1];
+                countryCode = splitLocale[2];
+            }
+
+            var supportedLanguageCode = languages.supportedCodes.map(function(locale) {
```
1ec5: This is an array, so the name should be plural.

index.js
```diff
@@ -243,6 +253,56 @@ module.exports = function(version, _options) {
...
+            // Same language code and country code (lng-CC)
+            } else if (languages.instructions[languageCode + '-' + countryCode]) {
+                return languageCode + '-' + countryCode;
+
+            // Same language code (lng)
+            } else if (supportedLanguageCode.indexOf(languageCode) > -1) {
+                return languages.supportedCodes[supportedLanguageCode.indexOf(languageCode)];
+
+            // Same language code and any script code (lng-Scpx)
+            } else if (supportedLanguageCode.indexOf(languageCode) > -1 && scriptCode) {
```
1ec5: If we support `zh-Hans` but not `zh`, and `language` is `zh-Hant`, then we won’t go into this `if` statement because `zh-Hant` isn’t one of the supported languages. Instead, we need to look for an item in `supportedLanguageCodes` that contains the language code `zh` and also contains a script code.

bsudekum: Addressed in https://github.com/Project-OSRM/osrm-text-instructions/pull/168/commits/167f7c2e783d043d6bce307f4173d16115b7f0f0

index.js
```diff
@@ -243,6 +246,58 @@ module.exports = function(version, _options) {
...
+        getBestMatchingLanguage: function(language) {
+            if (languages.instructions[language]) return language;
+
+            // If the language is not found, try a little harder
+            var splitLocale = language.toLowerCase().split('-');
+            var languageCode = false;
+            var scriptCode = false;
+            var countryCode = false;
+
+            if (splitLocale.lenth === 1) {
```
1ec5: Typo: length.

index.js
```diff
@@ -243,6 +246,58 @@ module.exports = function(version, _options) {
...
+                languageCode = splitLocale[0];
```
1ec5: Might as well set `languageCode` unconditionally at this point. If `language` is something bogus like `c++`, then we’ll end up falling back to English anyways.

index.js
```diff
@@ -243,6 +246,58 @@ module.exports = function(version, _options) {
...
+            } else if (splitLocale.length === 2 && splitLocale[1].length === 4) {
+                languageCode = splitLocale[0];
+                scriptCode = splitLocale[1];
+            } else if (splitLocale.length === 2 && splitLocale[1].length === 2) {
+                languageCode = splitLocale[0];
+                countryCode = splitLocale[1];
+            } else if (splitLocale.length === 3) {
+                languageCode = splitLocale[0];
+                scriptCode = splitLocale[1];
+                countryCode = splitLocale[2];
```
1ec5: Here’s [a useful reference](https://en.wikipedia.org/wiki/IETF_language_tag#Syntax_of_language_tags) on the syntax of a BCP47 language tag. We’re currently handling all the parts that a client is likely to pass in.

index.js
```diff
@@ -243,6 +246,58 @@ module.exports = function(version, _options) {
...
+                return languageCode + '-' + countryCode;
+
+            // Same language code (lng)
+            } else if (languages.instructions[languageCode]) {
+                return languageCode;
+
+            // Same language code and any script code (lng-Scpx) and the found language contains a script
+            } else if (supportedLanguageCodes.indexOf(languageCode) > -1 && scriptCode &&
+                languages.supportedCodes[supportedLanguageCodes.indexOf(languageCode)].split('-').length > 1 &&
+                languages.supportedCodes[supportedLanguageCodes.indexOf(languageCode)].split('-')[1].length === 4) {
```
1ec5: This would be easier if, right off the bat, we map each item in `supportedCodes` to an object with the keys `language`, `script`, and `region`.

index.js
bsudekum: I'm going to keep the splitting code and the return code separated.

index.js
bsudekum: It'd only be used really for this case. I'm fine with keeping this.

1ec5: If we do the preprocessing suggested above, then this becomes:

```js
else if (availableLanguages.find(function (language) {
  return language.language === languageCode && language.script;
})
```

and some of the duplication in the next `else if` block goes away.

We can do that preprocessing inside languages.js to avoid any meaningful performance penalty.

bsudekum: Changed it around to this in https://github.com/Project-OSRM/osrm-text-instructions/pull/168/commits/9a19e1eacde4f7bbdc48543c1f26b49476ce284a, not sure if it's any more clear.

index.js
```diff
@@ -243,6 +246,79 @@ module.exports = function(version, _options) {
...
+            // Same language code and any country code (lng-CX)
+            } else if (supportedLanguageCodes.indexOf(languageCode) > -1 && countryCode) {
+                return languages.supportedCodes[supportedLanguageCodes.indexOf(languageCode)];
+            } else {
+                return 'en';
+            }
+        },
+        divideLanguageCodes: function(language) {
+            // If the language is not found, try a little harder
+            var splitLocale = language.toLowerCase().split('-');
```
1ec5: To me, it’d be cleaner to use a regular expression for this job:

```js
// Two-letter language code, then optional four-letter script code, then optional two-letter region code
var match = language.match(/(\w\w)(?:-(\w\w\w\w))?(?:-(\w\w))?/i);
return {
  locale: match[0],
  language: match[1],
  script: match[2],
  region: match[3]
};
```

test/index_test.js
```diff
@@ -190,12 +190,98 @@ tape.test('v5 compile', function(t) {
...
+        t.assert(compiler('v5').getBestMatchingLanguage('foo'), 'en');
+        t.assert(compiler('v5').getBestMatchingLanguage('en-US'), 'en');
+        t.assert(compiler('v5').getBestMatchingLanguage('zh-CN'), 'zh-Hans');
+        t.assert(compiler('v5').getBestMatchingLanguage('zh-Hant'), 'zh-Hans');
+        t.assert(compiler('v5').getBestMatchingLanguage('zh-Hant-TW'), 'zh-Hans');
+        t.assert(compiler('v5').getBestMatchingLanguage('zh'), 'zh-Hans');
+        t.assert(compiler('v5').getBestMatchingLanguage('es-MX'), 'es');
+        t.assert(compiler('v5').getBestMatchingLanguage('es-es'), 'es-es');
+        t.end();
+    });
```
1ec5: Let’s add some tests of `divideLanguageCodes()` or whatever we call it once we move it to languages.js.

index.js
bsudekum: This is failing for me. I think both styles of parsing out the codes is fine.

index.js
```diff
@@ -243,6 +246,79 @@ module.exports = function(version, _options) {
...
+            // Same language code (lng)
+            } else if (languages.instructions[languageCode]) {
+                return languageCode;
+            // Same language code and any script code (lng-Scpx) and the found language contains a script
+            } else if (availableLanguages.find(function (language) {
+                return language.languageCode === languageCode && language.scriptCode;
+            })) {
+                return languages.supportedCodes[supportedLanguageCodes.indexOf(languageCode)];
+            // Same language code and any country code (lng-CX)
+            } else if (supportedLanguageCodes.indexOf(languageCode) > -1 && countryCode) {
```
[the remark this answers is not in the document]
bsudekum: added here: https://github.com/Project-OSRM/osrm-text-instructions/pull/168/files#diff-979945c097652b2859f6887cda2f7178R282

test/index_test.js
bsudekum: Added here: https://github.com/Project-OSRM/osrm-text-instructions/pull/168/files#diff-979945c097652b2859f6887cda2f7178R288

index.js
1ec5: If it’s failing it’s probably because the rest of the code hasn’t been updated to account for the normalized case being used by ☝️.

index.js
```diff
@@ -243,6 +246,49 @@ module.exports = function(version, _options) {
...
+            // Same language code (lng)
+            } else if (supportedLanguageCodes[languageCode]) {
+                return languageCode;
+            // Same language code and any script code (lng-Scpx) and the found language contains a script
+            } else if (languages.parsedSupportedCodes.find(function (language) {
+                return language.language === languageCode && language.scriptCode;
+            })) {
+                return languages.parsedSupportedCodes.find(function (language) {
+                    return language.languageCode === languageCode && language.scriptCode;
+                }).locale;
```
1ec5: Instead of repeating the `find()` call, store the result of the first find. It’d be a lot easier to store things in local variables if you replace all the `else if`s with `if`s – after all, we’re returning early in each case.

test/index_test.js
```diff
@@ -190,12 +190,116 @@ tape.test('v5 compile', function(t) {
...
+        t.deepEqual(languages.parseLanguageIntoCodes('en-US'), { countryCode: 'us', languageCode: 'en', locale: 'en-US', scriptCode: false });
+        t.deepEqual(languages.parseLanguageIntoCodes('zh-CN'), { countryCode: 'cn', languageCode: 'zh', locale: 'zh-CN', scriptCode: false });
+        t.deepEqual(languages.parseLanguageIntoCodes('zh-Hant'),  { countryCode: false, languageCode: 'zh', locale: 'zh-Hant', scriptCode: 'hant' });
+        t.deepEqual(languages.parseLanguageIntoCodes('zh-Hant-TW'),  { countryCode: 'tw', languageCode: 'zh', locale: 'zh-Hant-TW', scriptCode: 'hant' });
+        t.deepEqual(languages.parseLanguageIntoCodes('zh'), { countryCode: false, languageCode: 'zh', locale: 'zh', scriptCode: false });
+        t.deepEqual(languages.parseLanguageIntoCodes('es-MX'), { countryCode: 'mx', languageCode: 'es', locale: 'es-MX', scriptCode: false });
+        t.deepEqual(languages.parseLanguageIntoCodes('es-ES'), { countryCode: 'es', languageCode: 'es', locale: 'es-ES', scriptCode: false });
+        t.deepEqual(languages.parseLanguageIntoCodes('pt-PT'),  { countryCode: 'pt', languageCode: 'pt', locale: 'pt-PT', scriptCode: false });
```
1ec5: The Directions API currently allows a `language` parameter to be specified in any case, such as `pt-pt` or `PT-Pt`. If the Directions API is to use `getBestMatchingLanguage()`, we should test that this method matches case insensitively, so `pt-pt` should also produce `pt-PT`.

languages.js
```diff
@@ -49,8 +49,49 @@ var grammars = {
...
+    } else if (splitLocale.length === 3) {
+        scriptCode = splitLocale[1];
+        countryCode = splitLocale[2];
+    }
+
+    return {
+        locale: language,
+        languageCode: languageCode,
+        scriptCode: scriptCode,
+        countryCode: countryCode
```
1ec5: It’s a bit redundant to say “code” in each of these keys.

PR #170, opened 2017-10-06
Accept callback for formatting tokens

### LINKED ISSUE #52: Callback for token replacement values in compile

1ec5:
Added a `formatToken` option to `compile`, `getWayName`, and `tokenize` that gives the client an opportunity to manipulate a road name or other token value after any transformations built into this library but before the value is inserted into the overall instruction string. This feature is based on a similar feature built into the Swift version of this library.

Serving suggestion:

```js
var instruction = osrmTextInstructions.compile(language, step, {
  formatToken: function (token, value) {
    // Mark up numbers to be pronounced like numbers in addresses.
    value = value.replace(/\w*[\d.-]+\w*/g, `<say-as interpret-as="address">${value}</say-as>`);
    // Wrap road names in their pronunciations, if available.
    switch (token) {
      case 'name': case 'way_name':
        return step.pronunciation ? `<phoneme ph="${step.pronunciation}">${value}</phoneme>` : value;
      case 'rotary_name':
        return step.rotary_pronunciation ? `<phoneme ph="${step.rotary_pronunciation}">${value}</phoneme>` : nil;
      default:
        return value;
    }
  }
});
```

Fixes #52.

/cc @mcwhittemore @allierowan @bsudekum

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,11 +2,7 @@ var languages = require('./languages');
 var instructions = languages.instructions;
 var grammars = languages.grammars;
 
-module.exports = function(version, _options) {
-    var opts = {};
-    opts.hooks = {};
-    opts.hooks.tokenizedInstruction = ((_options || {}).hooks || {}).tokenizedInstruction;
-
+module.exports = function(version) {
     Object.keys(instructions).forEach(function(code) {
         if (!instructions[code][version]) { throw 'invalid version ' + version + ': ' + code + ' not supported'; }
     });
@@ -182,10 +178,6 @@ module.exports = function(version, _options) {
                 instruction = instructionObject.name;
             } else {
                 instruction = instructionObject.default;
-            }
-
-            if (opts.hooks.tokenizedInstruction) {
-                instruction = opts.hooks.tokenizedInstruction(instruction);
             }
 
             var nthWaypoint = options && options.legIndex >= 0 && options.legIndex !== options.legCount - 1 ? this.ordinalize(language, options.legIndex + 1) : '';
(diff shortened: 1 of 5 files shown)


1ec5:
This feature enables a superset of the use cases addressed by the `tokenizedInstruction` hook that wasy added in #79, although this feature is a little more verbose. Should we remove `tokenizedInstruction`?

index.js
```diff
@@ -226,22 +226,36 @@ module.exports = function(version, _options) {
...
             if (!language) throw new Error('No language code provided');
             // Keep this function context to use in inline function below (no arrow functions in ES4)
             var that = this;
-            var output = instruction.replace(/\{(\w+):?(\w+)?\}/g, function(token, tag, grammar) {
-                var name = tokens[tag];
-                if (typeof name !== 'undefined') {
-                    return that.grammarize(language, name, grammar);
-                }
+            var startedWithToken = false;
+            var output = instruction.replace(/\{(\w+)(?::(\w+))?\}/g, function(token, tag, grammar, offset) {
```
yuryleb: Wow, non-capturing group :+1:

Issue #155, opened 2017-09-14
Provide english fallback via `index.js` rather than via the `{lang}.json` files


mcwhittemore:
Currently when new strings are added to `en.json` we have to add copies of these english values to the other `{lang}.json` files. This is so something is provided in the new case before even if a translation is not provided.

Rather then doing this in the `{lang}.json` files we should provide fallback functionality either in `languages.js` or as part of `compile` in `index.js`.

PR #177, opened 2017-10-17
add depart with distance

### LINKED ISSUE #172: Add a "Head {direction} on {way_name} for {distance}" phrase
When giving a direction like `Continue on 16th street` or `Head south on St James Boulevard` it would be nice to optionally add on the distance. We can currently do this for the [continue](https://github.com/Project-OSRM/osrm-text-instructions/blob/7baf2ec6c9770f65bdd8b39f21cbf4caa88b1553/languages/translations/en.json#L97) but it would be nice to be able do this departures as well.

lyzidiamond:
Closes #172.

## Tasklist

 - [x] Add translation entry in `en.js`
 - [x] Add tests for continue with distance and depart with distance
 - [x] Add changelog entry
 - [x] Review

cc @mcwhittemore

### DIFF
--- a/languages/translations/en.json
+++ b/languages/translations/en.json
@@ -125,7 +125,8 @@
         "depart": {
             "default": {
                 "default": "Head {direction}",
-                "name": "Head {direction} on {way_name}"
+                "name": "Head {direction} on {way_name}",
+                "namedistance": "Head {direction} on {way_name} for {distance}"
             }
         },
         "end of road": {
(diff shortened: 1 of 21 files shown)


languages/translations/en.json
```diff
@@ -125,7 +125,8 @@
         "depart": {
             "default": {
                 "default": "Head {direction}",
-                "name": "Head {direction} on {way_name}"
+                "name": "Head {direction} on {way_name}",
+                "namedistance": "Head {direction} on {way_name} for {distance}"
```
1ec5: In the absence of #155, do we need to also add this line to all the other translation files?

mcwhittemore: hmmm... if we're going to do that, I'd like to use `depart name` rather than the english here.

lyzidiamond: I did this!

lyzidiamond:
Okay @mcwhittemore one last review and then we should be able to merge here.

PR #179, opened 2017-10-19
Update en.json to use traffic circle for all circles

mcwhittemore:
I haven't updated the 'roundabout turn' instructions here as I'm not sure 'traffic circle' is the best option in this case.

While 'rotary' and 'roundabout' are more precise, 'traffic circle' is more largely understood, at least in US English.

@1ec5 do you have any thoughts on this or insight into the `roundabout turn` situation?

### DIFF
--- a/languages/translations/en.json
+++ b/languages/translations/en.json
@@ -350,9 +350,9 @@
         "rotary": {
             "default": {
                 "default": {
-                    "default": "Enter the rotary",
-                    "name": "Enter the rotary and exit onto {way_name}",
-                    "destination": "Enter the rotary and exit towards {destination}"
+                    "default": "Enter the traffic circle",
+                    "name": "Enter the traffic circle and exit onto {way_name}",
+                    "destination": "Enter the traffic circle and exit towards {destination}"
                 },
                 "name": {
                     "default": "Enter {rotary_name}",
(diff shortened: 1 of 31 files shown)


1ec5:
> While 'rotary' and 'roundabout' are more precise, 'traffic circle' is more largely understood, at least in US English.

This is regional, of course. In recent years, roundabouts have been introduced to the Midwest in large numbers as "roundabouts", whereas they were known primarily as "traffic circles" when they were much rarer there. My sense is that "roundabout" is also more familiar on the West Coast. But the term "traffic circle" is pretty self-explanatory, so it isn't a huge deal either way.

> do you have any thoughts on this or insight into the roundabout turn situation?

A "roundabout turn" is just a turn at a normal intersection that happens to have an island in the middle as a traffic calming device. We call it out for informational purposes, similar to "at the end of the road" or "at the fork". I've never heard of this traffic calming device referred to as a traffic circle, only as a roundabout.

mcwhittemore:
Yea, I feel like your right @1ec5. For reference, here is a `roundabout turn`. Its to big to say 'turn left' but it seems weird to call something like this a traffic circle.

image

https://www.mapbox.com/get-directions/#20.72/52.51589/13.31471?coordinates=13.314773475857265,52.51582658410774;13.314641210543584,52.51595465673006

1ec5:
Interesting, that's a bit larger than the traffic calming device that I originally reported that led to the introduction of that maneuver type: Project-OSRM/osrm-backend#2718.

mcwhittemore:
> Since "traffic circle" isn't used in the U.K., where they're exceedingly common, this is a good argument in favor of finally adding an en-GB localization.

Yea, I was wondering that.

daniel-j-h:
> Since "traffic circle" isn't used in the U.K., where they're exceedingly common, this is a good argument in favor of finally adding an en-GB localization.

cc @systemed :P

systemed:
Yes, definitely roundabout in the UK, though in the West Midlands they're known as [islands](http://www.openstreetmap.org/way/145344979), and in Leicester as [circles](http://www.openstreetmap.org/way/193811700) sometimes. Plus we have [gyratories](http://www.openstreetmap.org/way/100175833), and if you really want to stress-test the directions, the [Magic Roundabout](http://www.openstreetmap.org/relation/7462227).

PR #181, opened 2017-10-23
Change default exit rotary verbiage

allierowan:
## Tasklist

 - [x] Add translation entry in each language file
 - [x] Add tests
 - [x] Add changelog entry
 - [ ] Review


mcwhittemore:
Hmmm... I'm not sure forcing all languages to adopt this verbiage is correct. That said, I'm not sure how we'd flag this phrase for translation otherwise. Meh. We need a better workflow for this kind of thing.

@allierowan - did you try only doing this to English first?

allierowan:
@mcwhittemore no, but I can do that. I thought this was the workflow I was supposed to follow - I suppose we don't know how this situation is handled in other languages. Happy to just do English if you think that makes more sense?

mcwhittemore:
Yea, lets just do English. This is the workflow when adding new content and it is maybe the right way to ask translators to reconsider their text, but while they do that it results in english being used for every language. This is one of my bigger pain points with this lib.

lyzidiamond:
@mcwhittemore why would we change the established workflow here?

> This is the workflow when adding new content and it is maybe the right way to ask translators to reconsider their text, but while they do that it results in english being used for every language.

This is true, but it is the current workflow and we should stick to it until there is a systemic change.

mcwhittemore:
> why would we change the established workflow here?

This PR is changing content rather than adding content. The workflow for changing content is to update the translation in transifex, but that is a bit weird with English as its the target languages for all other languages. Because its weird, the way we've done changes like this before is to only change english, trusting that the other languages make sense (which they very well might not).

As a bit of an aside, for 1.x this has to be one of the key things we change. Tightly coupling all translations to the english phrasing doesn't make sense.

lyzidiamond:
@mcwhittemore +1, I'm with you here!

Issue #61, opened 2016-10-31
How to deal with Arrivals?


freenerd:
Currently we have one step in OSRM for the arrival. We are emitting one instruction for it:

```js
{
    "step": {
        "maneuver": {
            "modifier": "left",
            "type": "arrive"
        },
        "name": "Street Name"
    },
    "instruction": "You have arrived at your destination, on the left"
}
```

To deal correctly with arrivals, we would actually need TWO instructions though, one after the last instruction before arrival, one on arrival:

- `In 200 meters, you will arrive at your destination, on the left`
- `You have arrived at your destination, on the left`

There are several solutions here:

1. The client could deal with this exclusively (given that they also already implement distances). The text instruction would need to be in the present then:
  - `In 200 meters, you arrive at your destination, on the left`
  - `You arrive at your destination, on the left`
2. We make osrm-text-instructions alert level aware
  - Needs a new ticket, i'm still undecided if and how we should do this
3. We add a new arrivals type in osrm-backend
  - This sounds like it would be way too much effort with something that is actually very specific to the layers above.

For now I'd go with option 1.

/cc @MoKob @daniel-j-h @TheMarex @danpat @1ec5 @willwhite @pveugen

1ec5:
For option 1, we should use the imperative like in every other instruction, not present tense. So: "Arrive at your final destination".

TheMarex:
Just a note on 3 because I think there is some context needed here on how we encode instructions:

- `maneuver` is the Maneuver that a driver needs to execute to turn on the current road segment
- everything else in `RouteStep` references the segment that follows.

That means for the a normal route we would see this a as the last two instructions:

```
[
    {
        "maneuver": {
            "modifier": "left",
            "type": "turn"
        },
        "duration": 50,
        "distance": 200,
        "name": "Street Name"
    },
    {
        "maneuver": {
            "modifier": "left",
            "type": "arrive"
        },
        "duration": 0,
        "distance": 0,
        "name": "Street Name"
    },
]
```

The information that the arrival will happen in 200m is already encoded in the previous instruction. Inserting a new instruction type between those instructions would break the convention that an `RouteStep` describes the route from the maneuver location until the location of the next maneuver.

Hence I agree with @freenerd's assessment that 3 would be a complicated addition to OSRM and IMHO probably more confusing then helping.

1ec5:
[...]

Issue #134, opened 2017-07-27
Arrival instruction should name the waypoint


1ec5:
For an arrival instruction, this library says “You have arrived at the destination”. Instead of a generic “the destination”, the instruction should say the waypoint’s name, like “You have arrived at White House”.

Unfortunately, OSRM generally gives the associated street as the waypoint’s name, which would cause the instruction to mislead, e.g., “You have arrived at Pennsylvania Avenue” even though the user has been traveling on Pennsylvania Avenue since the beginning of the previous step. Fortunately, the developer can modify the waypoint’s name before passing the route step object into this library. (This is also true in Swift: MapboxDirections.swift persists the waypoint name provided in the original request.)

While we’re at it, it would be nice to change the tense from the present perfect (“You have arrived at”) to the imperative (“Arrive at”): #61.

/ref mapbox/mapbox-navigation-ios#419 mapbox/mapbox-navigation-ios#422
/cc @bsudekum

1ec5:
To be clear, the step’s name is the waypoint name, so we won’t need to introduce any additional inputs.

hdaymon:
Need to verify how we want to do this.

1ec5:
> Fortunately, the developer can modify the waypoint’s name before passing the route step object into this library.

Note that the developer can technically get the name from any source they want, whether OpenStreetMap, the Mapbox Geocoding API, or a hard-coded string literal.

hdaymon:
From @1ec5 

> this is something that is very doable as long as OSRMTI runs on the client side. If we do this on the server side, then api-directions needs to do some postprocessing of what it gets from OSRM. Specifically, it needs to ignore whatever OSRM names the waypoint in favor of what it was named in the URL query parameter.
> 
> [11:32] 
> But that query parameter’s value comes from our Geocoding API in the case of Voyage and other clients.
> 
> [11:33] 
> This treads kind of close to the discussion about adding geocoding capabilities to the Directions API, which has always ended in :no_good:. That’s a non-issue when everything stays on the client side, b/c the POI name gets inserted at the last possible moment, right before sending off to the text-to-speech engine. 
>

1ec5:
Fixed in #184.

Issue #110, opened 2017-06-07
Split arrival instruction into counting and non-counting


patjouk:
Reviewing the French translation, I saw one thing that doesn't sound French but I couldn't fix it:

`You have arrived at your destination`: there's no way to remove "votre (your)" in the French translation [right now](https://github.com/Project-OSRM/osrm-text-instructions/blob/master/languages/translations/en.json#L53-L78). You would say "Vous êtes arrivés à votre seconde destination (you arrived at your second destination)" but  "Vous êtes arrivés à destination (you arrived at your destination)".

One solution could be to have a distinct string for 'You arrived", but without the number.

yuryleb:
This is also good for Russian :wink: It's really better to have one string for main/final destination and another for all intermediate ones.

freenerd:
Splitting counting arrival instructions and non-counting instructions into separate translatable strings makes sense to me.

The design may be like this then:

```json
{
        "arrive": {
            "default": {
                "default": "You have arrived at your destination"
            }
        },
        "arrive_counting": {
            "default": {
                "default": "You have arrived at your {nth} destination"
            }
     }
}
```

I'll change the title of this ticket accordingly.

@bsudekum is this something you'd want to look into as part of #111?

bsudekum:
@freenerd let's split this out from #111

patjouk:
Closing this issue. Please, create another one if you want to work on this.

PR #184, opened 2017-10-25
Adds an instruction worded for upcoming arrival

allierowan:
- [x] Add phrases in translation files
 - [x] Add changelog entry
 - [x] Add tests
 - [ ] Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log
 All notable changes to this project will be documented in this file. For change log formatting, see http://keepachangelog.com/
 
-## 0.10.4 2017-10-25
+## 0.10.4 2017-10-26
 
 - Add a phrase for an upcoming arrival step [#184](https://github.com/Project-OSRM/osrm-text-instructions/pull/184)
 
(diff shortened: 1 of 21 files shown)


1ec5:
This fixes #61.

1ec5:
Technically the Directions API can at any time produce an unexpected `modifier` – hence the `default` cases. In that case, `compile` will wind up with a more deeply nested object than expected:

https://github.com/Project-OSRM/osrm-text-instructions/blob/d91209a433087350a7782bb5bb176c52244ed36b/index.js#L129-L137

We have special cases for `use lane`, `rotary`, and `roundabout` instructions that dig deeper into the object, so that `instruction` ends up being a string rather than an object. We’ll need a similar case for `arrive`. Our tests aren’t catching this case because they only cover the known `modifier`s.

By the way, we’ll eventually need more keys alongside `default` for #110 and #134.

1ec5:
> Technically the Directions API can at any time produce an unexpected `modifier` – hence the `default` cases. In that case, `compile` will wind up with a more deeply nested object than expected

My bad, I was misreading the PR. It’s odd that additions to the `v5` object don’t result in any changes to index.js, but I take it that’s because these strings are intended to be used in `tokenize` only and not in `compile`, correct?

allierowan:
@1ec5 yeah, that was what I was thinking. I suppose we could expose it to compile by adding some sort of flag?

PR #186, opened 2017-10-26
Added new untranslated Swedish strings related to roundabouts and rotary

davols:
Added untranslated strings in transifex, pulled them and created the test strings. 

cc @frederoni for review and @1ec5 

Started wondering about the possibility with commas and how they are treated with TTS in Swedish. Like `i rondellen, ta av mot` vs `i rondellen ta av mot`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file. For change 
 ## master
 
 - Added a `merge straight` instruction [#183](https://github.com/Project-OSRM/osrm-text-instructions/pull/183)
+- Added new untranslated Swedish strings related to roundabouts and rotary [#186](https://github.com/Project-OSRM/osrm-text-instructions/pull/186)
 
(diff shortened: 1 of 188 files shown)


frederoni:
The comma adds ~300ms wait (using Polly TTS)  which sounds much better in my opinion.
The voice becomes more robotic w/o the comma.

screenshot 2017-10-26 15 14 21

davols:
> The comma adds ~300ms wait (using Polly TTS) which sounds much better in my opinion. The voice becomes more robotic w/o the comma.

Sounds like a good idea with the comma in that case. I haven't listened or tested it myself yet.

There are some strings that should have a comma already. I could update this PR with commas on the new strings and also the ones that should have it.

davols:
Just tried with Polly as well, it does sound way better. Updated the PR with the new `arrive.upcoming` strings. 

Got a merge conflict which I think is solved, the PR looks to have the latest stuff. Please have a look again @frederoni

PR #189, opened 2017-10-28
Synchronize translations with Transifex

1ec5:
Pulled and selectively applied the latest translations in Transifex, excluding any changes that would revert a translated string back to English.

/cc @allierowan @mcwhittemore


1ec5:
These translations were contributed by:

* @Verdulo in Esperanto
* @sealos and me in Spanish
* @paillou in French
* @gloffreda in Portuguese
* @yuryleb in Russian

1ec5:
> But I missed when `arrive.<maneuver>.upcoming` appeared - when is it supposed to be used, before/instead of `arrive.<maneuver>.default` or for via points only?

This instruction was added in #184. It isn’t used by `compile()`, but a client can use it with `tokenize()` to come up with an instruction like this:

> Turn left, then, in 500 feet, you will arrive at your destination, on the left.

yuryleb:
Thank you, then I should correct translation again to use more appropriate tense. Please synchronize Russian with Transifex more once (some other `arrival` phrases are also extended).

mcwhittemore:
Rebasing vs master.

languages/translations/eo.json
```diff
@@ -422,53 +422,23 @@
...
-                "name": "Veturu rekten al {way_name}",
-                "destination": "Veturu rekten direkte al {destination}"
+                "default": "Eliru trafikcirklegon",
+                "name": "Elveturu trafikcirklegon al {way_name}",
+                "destination": "Elveturu trafikcirklegon direkte al {destination}"
             }
         },
         "turn": {
             "default": {
-                "default": "Veturu {modifier}",
```
mcwhittemore: This changes seems suspect to me. Can someone confirm that there is no verb needed to tell someone to turn in esperanto?

mcwhittemore: Its also possible that the direction word is its own verb.

1ec5: In this localization, the `modifier` words are all translated into adverbs like “maldekstren” (left), which can’t technically be used on their own as imperative verbs. Colloquially, one can say “Left!” in English, and I’d imagine that’s true in most other languages too, including Esperanto. But I think the previous translation was more faithful to the original English. Perhaps @tcql can weigh in.

@verdulo originally put in “Veturu” (to go), which I think makes sense here, but recently removed it, perhaps because one of the direction words, `uturn`, was translated as “turniĝu malantaŭen”, so “Veturu turniĝu malantaŭen” would’ve been redundant. However, this particular set of strings (`turn.default.*`) is never used, because all the possible `modifier`s under `turn` have specific translations.

https://github.com/Project-OSRM/osrm-text-instructions/blob/5f78287907cf83c1495cc3a45f64fefe600b1dc8/test/fixtures/v5/turn/left_default.json#L12

Nor does the `uturn` modifier translation ever get used, because the `turn.uturn.*` strings don’t include `{modifier}`.

https://github.com/Project-OSRM/osrm-text-instructions/blob/5f78287907cf83c1495cc3a45f64fefe600b1dc8/test/fixtures/v5/turn/uturn_default.json#L12

For now, I’ve reverted that change in Transifex and on this branch.

1ec5:
Published v0.10.6.

1ec5:
Thanks @yuryleb – removed in #191.

PR #193, opened 2017-11-22
Call formatToken on bare refs and names

1ec5:
When the step’s way has only a ref or a name, call `formatToken` to give the client the opportunity to manipulate that ref or name, just as it would if the way has both a ref and a name.

/cc @mcwhittemore


1ec5:
> Can we get some tests so that we know this doesn't get broken in the future.

Added in 65a28cdce1b0f43ca46f6482a089572b50d6e42b.

PR #195, opened 2017-11-27
add info about adding new instruction options

### LINKED ISSUE #185: Document adding new language features without introducing english into other languages
Sometimes we add something to OSRM Text Instructions that requires us to add a new string to all of the translation files. When we do this, we should set the initial value of this new feature the text of a very similar feature in that language. This is a change from how we've been doing things, where we set the new feature to English.

mcwhittemore:
Fixes #185


mcwhittemore:
Yea, I agree @lyzidiamond. I thought about starting this and then thought about my desire to change how the translations files work and opting for the quick solution so we don't throw away work later (maybe).

PR #197, opened 2017-12-07
Consistently choose the first destination ref and destination

### LINKED ISSUE #135: Destinations with refs are parsed inconsistently
[This line](https://github.com/Project-OSRM/osrm-text-instructions/blob/bdf31a49565e76db76a338f05913087c2d99b9db/index.js#L185) assumes that a step’s `destination` is a comma-separated list, but it can also take the form `Ref1, Ref2: Destination1, Destination2`. So there’s inconsistent behavior around destinations:

* [`I 71 North, US 50 East: Columbus`](http://map.project-osrm.org/?z=16&center=39.100176%2C-84.515848&loc=39.109551%2C-84.532821&loc=39.098111%2C-84.513016&hl=en&alt=0) produces “towards I 71 North” but should produce “towards I 71 North: Columbus”
* [`I 471 South: Newport, KY`](http://map.project-osrm.org/?z=17&center=39.102374%2C-84.500570&loc=39.098261%2C-84.509604&loc=39.101791%2C-84.497437&hl=en&alt=0) produces “towards I 471 South: Newport” (but see Project-OSRM/osrm-backend#3329)

/ref Project-OSRM/osrm-text-instructions.swift#31
/cc @bsudekum

1ec5:
The logic for selecting the first destination was too simplistic, failing to account for the `Ref: Destination` format in the `destination` property.

Fixes #135.

/cc @lyzidiamond

### DIFF
--- a/index.js
+++ b/index.js
@@ -180,13 +180,23 @@ module.exports = function(version) {
                 instruction = instructionObject.default;
             }
 
+            var destinations = step.destinations && step.destinations.split(': ');
+            var destinationRef = destinations && destinations[0].split(',')[0];
+            var destination = destinations && destinations[1] && destinations[1].split(',')[0];
+            var firstDestination;
+            if (destination && destinationRef) {
+                firstDestination = destinationRef + ': ' + destination;
+            } else {
+                firstDestination = destinationRef || destination || '';
+            }
+
             var nthWaypoint = options && options.legIndex >= 0 && options.legIndex !== options.legCount - 1 ? this.ordinalize(language, options.legIndex + 1) : '';
 
             // Replace tokens
             // NOOP if they don't exist
             var replaceTokens = {
                 'way_name': wayName,
-                'destination': (step.destinations || '').split(',')[0],
+                'destination': firstDestination,
                 'exit': (step.exits || '').split(';')[0],
                 'exit_number': this.ordinalize(language, step.maneuver.exit || 1),
                 'rotary_name': step.rotary_name,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to this project will be documented in this file. For change log formatting, see http://keepachangelog.com/
 
+## master
+
+- Fixed inconsistent destinations when a ramp leads to multiple routes towards multiple places. [#197](https://github.com/Project-OSRM/osrm-text-instructions/pull/19y)
+
 ## 0.10.7 2017-12-05
 
 - Fixed issue preventing `formatToken` from being called on refs or names when the way has only a ref or a name. [#193](https://github.com/Project-OSRM/osrm-text-instructions/pull/193)
--- a/test/fixtures/v5/other/way_name_destination_refs.json
+++ b/test/fixtures/v5/other/way_name_destination_refs.json
@@ -1,0 +1,29 @@
+{
+    "step": {
+        "maneuver": {
+            "type": "off ramp",
+            "modifier": "right"
+        },
+        "destinations": "Ref1, Ref2: Destination 1, Destination 2"
+    },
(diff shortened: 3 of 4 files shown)


lyzidiamond:
This PR assumes that this will always take the format `ref1, ref2: dest1, dest2` and that is not always the case. In your own examples in #135, it's impossible to know that `ref1` matches `dest1` and `ref2` matches `dest2` -- it's just as likely that the refs are cosigned and the destinations, although they have two values, refer to the same ultimate destination. Or, in your first example there, `ref1, ref2: dest1` -- it's impossible to know if `dest1` only is matched with `ref1` and not `ref2`, and if we assume that, then does `ref2` not have a `destination`? Similarly, this PR doesn't address the issue you outlined in that ticket, where `Newport, KY` was splitting into `dest1, dest2` on the comma. Per planning with @mapbox/guidance yesterday, we decided to punt on #135 and maybe address it as a data issue instead of a parsing issue.

1ec5:
I’m remain aware of that example, but I would contend that this change is still an improvement. Better to say `I 71 North: Columbus` than just `I 71 North`, even if I-71 and I-75 are cosigned. We’ll continue to say `I 471 South: Newport`; that isn’t ideal, but it’s OK for now.

lyzidiamond:
We came to a conclusion here in chat, and I want to document it:

This PR's changes will primarily target ramps heading onto highways. Ideally, the ramps should have a `destination` value to indicate where the ramp is going -- if it's an onramp to I-84 East towards The Dalles, the `destination` value should be `I-84 East: The Dalles`. The announcement would be "Take the ramp on the right towards I-84 East: The Dalles".

In some situations, however, ramps will lead to multiple highways or a co-signed highway. In those cases, the `destination` value can look like this: `I-84 East, I-205 South: The Dalles, Salem`. Right now, the code assumes that `destination` will be a comma-separated list of destinations, which means the code will evaluate `destination` as having 3 values: `I-84 East`, `I-205 South: The Dalles`, and `Salem`. It will then return a phrase that only includes the first destination value: "Take the ramp on the right towards I-84 East."

This PR adds a bit more parsing and evaluating -- instead of blindly breaking `destination` apart on commas, @1ec5's code here accommodates this potential structuring by reformatting slightly. Once this PR is merged, `I-84 East, I-205 South: The Dalles, Salem` will be evaluated as `I-84 East: The Dalles, I-205 South: Salem`.

While there could be hiccups with this, presumably a ramp that leads to multiple highways would have a maneuver soon after that directs the driver onto one or the other. I think this is a good bug fix.

coxchapman:
Great writeup @lyzidiamond! Your explanation makes it really clear. Thanks @1ec5 for implementing.

index.js
```diff
@@ -180,13 +180,23 @@ module.exports = function(version) {
                 instruction = instructionObject.default;
             }
 
+            var destinations = step.destinations && step.destinations.split(': ');
```
lyzidiamond: I assume we can always split on `': '` because this value is generated from OSRM, yes?

1ec5: Yes, `': '` is inserted by OSRM:

https://github.com/Project-OSRM/osrm-backend/blob/fd57c5b48bd289ecf15ffa7ea703f0461536f784/profiles/lib/destination.lua#L23

For the other calls to `split()`, I passed in `','` instead of `', '`, because `tokenize()` already trims whitespace from each token:

https://github.com/Project-OSRM/osrm-text-instructions/blob/ad11173324c91e30a014ab61c23bb5043d130236/index.js#L215

A colon is also used [to mark a possessive or contraction](https://en.wikipedia.org/wiki/Colon_%28punctuation%29#Word-medial_separator) in Finnish and Swedish, so I included a space in `': '` just in case.

lyzidiamond: 👍

Issue #198, opened 2017-12-12
Omit exit side when it's the same as driving side


lyzidiamond:
Exit instructions should omit the exit side unless it is different from the driving side.

Currently (in the US, where `driving_side === 'right'`: "Take exit 5 on the right towards Main Street"
After this change: "Take exit 5 towards Main Street"
Currently: "Take exit 7 on the left towards First Ave"
After this change: "Take exit 7 on the left towards First Ave"

@1ec5 @hdaymon @mcwhittemore @allierowan @coxchapman

1ec5:
We previously attempted to remove the exit side unconditionally in #124 but rolled it back in in #125 due to adverse effects on wrong-sided exits. #199 once again removes the exit side, except for on wrong-sided exits.

lyzidiamond:
Closed in #199!

PR #199, opened 2017-12-12
Add omit side condition to compile

lyzidiamond:
# Issue

Ref #198. ~I would love a code review here, because what I did feels weird.~ WHAT I DID IS AWESOME

## Tasklist

 - [x] Add condition to `compile`
 - [x] Add tests
 - [x] Add changelog entry
 - [ ] Review

cc @1ec5 @mcwhittemore @allierowan


index.js
```diff
@@ -113,6 +113,7 @@ module.exports = function(version) {
             var type = step.maneuver.type;
             var modifier = step.maneuver.modifier;
             var mode = step.mode;
+            var side = step.driving_side;
```
1ec5: If the input comes from OSRM v5.0–v5.13, `driving_side` will be undefined. [`String.indexOf()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf) apparently returns −1 when given no argument, so `omitSide` ends up false, preserving the previous behavior. As long as we can rely on this behavior of `indexOf()`, it should be no problem, but can would you mind including a comment to this affect?

lyzidiamond:
Tests are added! Would love eyes on this from @allierowan and/or @mcwhittemore before merge.

index.js
```diff
@@ -130,10 +132,16 @@ module.exports = function(version) {
             var instructionObject;
             if (instructions[language][version].modes[mode]) {
                 instructionObject = instructions[language][version].modes[mode];
-            } else if (instructions[language][version][type][modifier]) {
-                instructionObject = instructions[language][version][type][modifier];
             } else {
-                instructionObject = instructions[language][version][type].default;
+              // omit side from off ramp if same as driving_side
+              // note: side will be undefined if the input is from OSRM <5.14
+              // but the condition should still evaluate properly regardless
```
1ec5: Weird that the linter complained about indentation below but not on these comment lines. 🤷‍♂️

PR #200, opened 2017-12-13
Update Romanian translation, use pt-BR for Portugese

ggl:
## Tasklist

 - [x] Update Romanian translation
 - [x] Use `pt-BR` for `pt` localization
 - [ ] Add changelog entry
 - [ ] Review

## Requirements / Relations


Rui-Santos:
I asked to join the pt-PT team at Transifex to finish the European Portuguese translation. In fact I already have it done. Translated the Json on my own for my use. Can submit a pull request here, or can submit it at Transifex. BTW, Indeed there are diferences. The grammar is mostly the same. The difference is only in a few expressions that Portuguese people don't use. Ex. "Pegue a rampa" becomes "Apanhe a rampa"(Take the ramp), "rotatória" becomes "rotunda"(roundabout).

Rui-Santos:
@1ec5 Well, I would resend the request if I could. Portuguese (Portugal) is not on the list anymore when I press the Join Team button. My Transifex username is psykickruhyn

Rui-Santos:
Just a quick question, are these only written instructions or are they also used for verbal instructions on turn-by-turn navigation apps/devices?

1ec5:
Both. The Mapbox Navigation SDK primarily uses them for verbal instructions, but other clients including Web-based clients typically display them visually.

Rui-Santos:
Ok, then I'm doing it well, using language which sounds more natural on spoken form. Was afraid I was rewriting too many of the verbal times of the previous translation. Still need to go back and change the structure of some sentences, also. I will open a new Issue because I have some doubts on othet stuff.

1ec5:
@Rui-Santos has kindly created an actual European Portuguese localization that we’re adding in #229. @ggl, unfortunately, this PR has become outdated, so please open a new PR whenever you’re ready to update the Romanian localization again. Thanks!

languages/translations/ro.json
```diff
@@ -92,37 +92,37 @@
...
+                "name": "Virați puternic la stânga pe {way_name}",
+                "destination": "Virați puternic la stânga spre {destination}"
             },
             "sharp right": {
-                "default": "Virați brusc dreapta",
-                "name": "Virați brusc stânga pe {way_name}",
-                "destination": "Virați brusc dreapta spre {destination}"
+                "default": "Virați puternic la dreapta",
+                "name": "Virați puternic la dreapta pe {way_name}",
+                "destination": "Virați puternic la dreapta spre {destination}"
```
1ec5: These translations need to be changed [at Transifex](https://www.transifex.com/project-osrm/osrm-text-instructions/) before they can be changed here. Otherwise, the changes would be overwritten the next time we update this project.

ggl: Is there any way to import the json into Transifex?

1ec5: There is, but it looks like @mdomnita modified some strings in the meantime, so it might be more prudent to merge the values by hand rather than overwriting what’s there. I’ve added you to the Romanian translation team so you can see what options are available. If you’re comfortable using a merge or diff tool on your computer, then you could export the strings, merge the changes, and reimport. But I’ll leave it up to you and @mdomnita as to the specific translations to use.

ggl:
@mdomnita is my colleague, his desk is in front of mine. I based my translation upon his initial work.
Thanx for adding me to the team, the remaining strings have been translated on Transifex.

1ec5:
I wound up including these translations in #231. Please take a look at that PR’s test fixtures to ensure that the generated sentences make sense. Thanks for your help, @ggl @mdomnita!

PR #207, opened 2018-01-12
Short arrival text

allierowan:
# Issue

We'd like a shorter arrival text option to save space- this adds a short option for both the upcoming and default arrival texts. For other languages, I copied over their regular default and upcoming phrases for now.

## Tasklist

 - [x] Add sentences for languages
 - [x] Add tests (added a test for upcoming too because it hadn't been tested)
 - [x] Add changelog entry
 - [x] Review


test/index_test.js
```diff
@@ -382,8 +382,9 @@ tape.test('v5 compile', function(t) {
                         options.legCount = fixture.options.legCount;
                         options.classes = fixture.options.classes;
                     }
-
                     Object.keys(fixture.instructions).forEach((l) => {
+                        // ignore custom instructions that don't get compiled
+                        if (!fixture.step) return;
```
mcwhittemore: What does this solve?

allierowan: this code was trying to test the new fixtures I added, but they don't get compiled by default, you have to specifically grab them by the right path. The fixtures I created don't have steps connected to them, so I just added this line so they don't get included

PR #208, opened 2018-01-18
Add Danish localization

1ec5:
Pulled a new Danish localization from Transifex.

@klavspc, thanks for the translations! When you get a chance, could you update the 22 strings that have become outdated over the past month? For now, this PR uses the English strings for those instructions, but I can update the PR as soon as the strings get translated in Transifex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file. For change 
 
 ## master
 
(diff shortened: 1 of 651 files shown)


1ec5:
The rest of the translations have been added in 1a46da8fbce345e7678bbf27fd64d2f788835af9.

1ec5:
The rest of the translations have been added in 1a46da8fbce345e7678bbf27fd64d2f788835af9.

/cc @allierowan

Issue #211, opened 2018-01-23
Use @mapbox/locale-utils


bsudekum:
The locale parsing logic has been abstracted into it's own library. It'd be a good idea to switch this over to have a single point of truth when it comes to parsing locales accurately.

When this switch is made, it might be a good idea to also implement https://github.com/Project-OSRM/osrm-text-instructions/issues/194

Link: https://github.com/mapbox/locale-utils

/cc @1ec5

PR #212, opened 2018-01-24
Remove getBestMatchingLanguage() and parseLanguageIntoCodes()

bsudekum:
# Issue

Closes: https://github.com/Project-OSRM/osrm-text-instructions/issues/211

This PR removes the functions `getBestMatchingLanguage()` and `parseLanguageIntoCodes()` in favor of abstracting the logic out in these functions [into their own library](https://github.com/mapbox/locale-utils). This came change about because these functions were being copied around into other projects to allow for a custom supported language list.

For the most part, it's a straight 1:1 copy between the code that was here and it's new home in https://github.com/mapbox/locale-utils. 

Noting neither one of the functions here were used internally in this repo. They were exposed as helper functions. 

/cc @allierowan @1ec5


Readme.md
```diff
@@ -27,12 +27,11 @@ Grammatical cases and other translated strings customization after [Transifex](h
 var version = 'v5';
 var osrmTextInstructions = require('osrm-text-instructions')(version);
 
-// If you’re unsure if the user’s locale is supported, use `getBestMatchingLanguage` method to find an appropriate language.
-var language = osrmTextInstructions.getBestMatchingLanguage('en-US');
+// If you’re unsure if the user’s locale is supported by OSRMTI, see `https://github.com/mapbox/locale-utils` for deciphering the best locale.
```
1ec5: Add another link outside the code block for convenience.

PR #217, opened 2018-02-05
Add "exit {exit}" to phrases

allierowan:
We are hardcoding the word "exit" because we don't have a phrase for it here. I have added it to the `phrases` section.
```
"phrases": {
              ...,
              "exit": "sortie {exit}"
            }
```
## Tasklist

 - [x] Add changelog entry
 - [ ] Review

@mcwhittemore @lyzidiamond @1ec5 thoughts on this approach? Should I add this section to all languages with the english fallback?


languages/translations/fr.json
```diff
@@ -41,6 +41,9 @@
                 "ox": "Serrer à gauche",
                 "xox": "Rester au milieu",
                 "oxo": "Rester à gauche ou à droite"
+            },
+            "vocabulary": {
+              "exit": "la sortie"
```
1ec5: Is this term being used as part of an instruction or in a standalone context, such as in banner text in the navigation SDKs? This library already incorporates the words “the exit” into some instruction formats:

https://github.com/Project-OSRM/osrm-text-instructions/blob/895a20330f017d6ad3ed6ec983dbcb1eba331ee2/languages/translations/fr.json#L299

allierowan:
per the discussion on https://github.com/mapbox/bumblebee/issues/243, I have made these updates.

cc @mcwhittemore @1ec5 @lyzidiamond

languages/translations/es.json
```diff
@@ -54,7 +54,8 @@
             "two linked by distance": "{instruction_one} y luego en {distance}, {instruction_two}",
             "two linked": "{instruction_one} y luego {instruction_two}",
             "one in distance": "A {distance}, {instruction_one}",
-            "name and ref": "{name} ({ref})"
+            "name and ref": "{name} ({ref})",
+            "exit with number": "la salida {exit}"
```
1ec5: The article _la_ should be omitted from this string. It wouldn’t appear in signage.

[](https://commons.wikimedia.org/wiki/File:PR-66_este_hacia_PR-3.jpg)

languages/translations/fr.json
```diff
@@ -54,7 +54,8 @@
             "two linked by distance": "{instruction_one}, puis, dans {distance}, {instruction_two}",
             "two linked": "{instruction_one}, puis {instruction_two}",
             "one in distance": "Dans {distance}, {instruction_one}",
-            "name and ref": "{name} ({ref})"
+            "name and ref": "{name} ({ref})",
+            "exit with number": "la sortie {exit}"
```
1ec5: France accompanies exit numbers with the [exit](https://commons.wikimedia.org/wiki/File:AB-AS.svg) icon instead of text, while Québec and New Brunswick display the number alone. But if these regions were to display text, it would omit _la_, as seen in this bilingual sign in New York:

[](https://commons.wikimedia.org/wiki/File:Northway_Exit_34_%28NY_915F%29.jpg)

languages/translations/es.json
allierowan: 👍

Issue #112, opened 2017-06-19
Correct german counting of multiple waypoints


freenerd:
As per https://github.com/Project-OSRM/osrm-text-instructions/pull/111/files?diff=split#r121345638 we need to fix the german translation of counting multiple waypoints:

Now: `Sie haben Ihr erste Ziel erreicht`.  
Correct: `Sie haben Ihr erstes Ziel erreicht`.

Options:

- See if we can find same-gender words to use.
- Make ordinalization gernder-aware. Not sure if that should be via tokens only then though.

1ec5:
The grammar feature added in #102 can be used to decline tokens for gender. Alternatively, we could avoid spelling out the ordinal numbers, translating “first” as “1.” I suspect most text-to-speech engines would be able to handle that representation correctly. Would it be OK to write it that way for display on screen?

/cc @daniel-j-h @yuryleb

yuryleb:
As I could see, `{nth}` is used only in [`arrive` block](https://github.com/Project-OSRM/osrm-text-instructions/blob/master/languages/translations/de.json#L59-L84), why don't correct [`ordinalize` constants](https://github.com/Project-OSRM/osrm-text-instructions/blob/master/languages/translations/de.json#L7-L18) right on Transifex with changing `erste` to `erstes` and so on? Russian translation also has these constants adopted to gender used in `arrive` instructions.

yuryleb:
Oh, sorry, currently these `ordinalize` contants are used as `{exit_number}` in `roundabout` instructions :disappointed: And they are adopted to `Ausfahrt` feminine gender. Actually I couldn't find appropriate synonym for the `Ziel` of the same gender.

So it's really it's possible to extended grammar with `masculine`, `feminine`, `neuter` options - for example `{exit_number:feminine}` and `{nth:neuter}`. I could prototype code/JSON for this but some German should check after me then.

daniel-j-h:
Usually you expand numbers up to twelve as words and only then you write them out as numbers.

In the context of an app with limit screen space I think it's totally fine to use numbers when the TTS engines can handle it anyway.

yuryleb:
Ok, no action from me yet :wink: 

But actually I supposed our instructions should be used by TTS primarily and be on screen for debug purposes only :wink: App should display just arrow icon(s) and way name/destination for the next maneuver, shouldn't it?

1ec5:
[...]

PR #220, opened 2018-02-14
Update German translation

suptuphup:
## Tasklist

 - [ ] Update German translation


yuryleb:
Maybe you could also fix German ordinals from "erste" to "1." as requested in #112 (on [Transifex](https://www.transifex.com/project-osrm/osrm-text-instructions/) first)?

PR #221, opened 2018-02-22
Add abbreviations

danpaz:
# Issue

Moves abbreviation lists into OSRMTI. For now, these abbreviations are simply exported as JSON with no additional logic integrating them into the library.

Usage: 
```
var osrmTextInstructions = require('osrm-text-instructions')('v5');
var abbreviations = osrmTextInstructions.abbreviations['en'];
```

## Tasklist

 - [X] Migrate abbreviation lists json into OSRMTI
 - [X] Export abbreviations as part of the module 
 - [x] Add changelog entry
 - [ ] Review

## Requirements / Relations

None.

cc @mcwhittemore @allierowan @lyzidiamond @1ec5


languages/abbreviations/bg.json
```diff
@@ -0,0 +1,68 @@
+{
+    "abbreviations": {
```
bsudekum: It might make more sense to make this key a little more descriptive since the folder name is already `abbreviations`.

yuryleb:
Maybe it's good idea to move map names abbreviations from MapBox SDK to more "public" place but how do you suppose to use them in text instructions? I hope you don't want to apply them to way names in instructions before output? They're mostly supposed to be used by TTS and abbreviations can only break pronunciation not to help instead. Also they can break or turn off grammar processing functionality (I spend a year or more for this :tired_face:)

Actually abbreviations are useful for map renders only IMHO. So maybe this good idea will land to some other repository?

languages/abbreviations/bg.json
1ec5: This set of abbreviations is a catch-all for miscellaneous words that don’t play a specific role in a road name or destination. I guess we could call it `miscellaneous`, but `abbreviations` is OK too in my opinion.

1ec5:
> Maybe it's good idea to move map names abbreviations from MapBox SDK to more "public" place but how do you suppose to use them in text instructions? I hope you don't want to apply them to way names in instructions before output? They're mostly supposed to be used by TTS and abbreviations can only break pronunciation not to help instead. Also they can break or turn off grammar processing functionality (I spend a year or more for this :tired_face:)

Not to worry; these abbreviations won’t be applied anywhere that a text-to-speech engine would see them. Although the sentences generated by this library are primarily fed to text-to-speech engines, they may also appear visually in a list of steps, such as on the Web.

At the moment, we’re planning to use these abbreviations in the turn banner in the navigation SDK, which is where they’re currently being used. There are already a few phrases and such in this repository that are left out of voice instructions, such as “exit”. If there are problems with housing them in the same library as voice instruction strings, then we can consider solutions to those problems separately from this PR.

languages/abbreviations/fr.json
```diff
@@ -0,0 +1,570 @@
...
+        "établissements": "Étabts",
+        "société anonyme": "SA",
+        "directeurs": "Dir.",
+        "vii": "7",
+        "culturel": "Cult.",
+        "central": "Ctral",
+        "métropolitaine": "Métrop.",
+        "administrations": "Admin.",
+        "amiraux": "Amir.",
+        "sur": "\/",
```
1ec5: These backslashes probably shouldn’t be in here, should they? If I’m not mistaken, “sur” would be “[s/](https://en.wiktionary.org/wiki/s/#French)”, not just “/”.

danpaz:
Thanks @1ec5 @lyzidiamond !

PR #228, opened 2018-03-16
Add RegExp 'global' option to Russian grammar rules

yuryleb:
# Issue

Some non-JS [string replace with regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) implementations, such as [Swift](https://github.com/Project-OSRM/osrm-text-instructions.swift/pull/49), always use [global search](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/global) option.

This PR adds this `g` option explicitly to Russian grammar meta block and updates tests to require `g` meta flags for all existing and future grammars.

## Tasklist

 - [x] Add explicit `g` option to grammar and tests
 - [ ] Add changelog entry
 - [ ] Test with osrm-frontend
 - [ ] Review


yuryleb:
osrm-frontend (with [formatToken() fix](https://github.com/Project-OSRM/osrm-frontend/pull/271)) also works.

PR #229, opened 2018-03-22
Added European Portuguese localization

1ec5:
# Issue

Added a European Portuguese localization from Transifex. Also updated the Brazilian Portuguese localization, which still has some gaps in translation.

@Rui-Santos, thanks for the translations! Please take a look at the test fixtures to ensure the generated sentences make sense. Feel free to keep refining the translations; we can update this PR in parallel until you think it’s ready for an initial release.

## Tasklist

 - [x] Update translations
 - [x] Add changelog entry
 - [ ] Review

## Requirements / Relations

This supersedes the Portuguese portion of #200. It complements the UI localizations added in mapbox/mapbox-navigation-ios#1252.

/cc @lyzidiamond @danpaz


1ec5:
Also, it would be good to add an [abbreviation file](https://github.com/Project-OSRM/osrm-text-instructions/tree/master/languages/abbreviations/) for Portuguese at some point, but that can happen in a separate PR.

test/fixtures/v5/exit_rotary/left_destination.json
```diff
@@ -21,7 +21,8 @@
         "it": "Svolta a sinistra verso Destination 1",
         "nl": "Ga linksaf richting Destination 1",
         "pl": "Skręć w lewo w kierunku Destination 1",
-        "pt-BR": "Vire à esquerda sentido Destination 1",
+        "pt-BR": "Exit the traffic circle towards Destination 1",
```
danpaz: Just double checking: the Brazilian Portugese translations are expected to be in English here because the translation is missing, right?

1ec5: Although it’s unfortunate that this string has reverted to English, the previous Portuguese string was suboptimal anyways (“Turn left towards Destination 1”). I suppose it would be possible to pare back this PR to bring back the old string and test fixture, but I’m not sure that should be the bar for adding a separate localization.

Rui-Santos:
Look fine to me. I guess any unnatural language quirks will only be noticed with it being used without the placeholders. I'll see if tomorrow I can put a couple hours aside and take care of the abbreviatons.

1ec5:
Sounds good. Feel free to open a PR or issue if you spot any problems or need any strings updated here.

Rui-Santos:
OK, I need more than a couple hours for the abbreviations. Need to parse a list with thousands of them and get only the ones that are used in toponymics. Take will take more time than I was expecting. It will take a few days. Will have to go through the list letter by letter, one at a time, when I get some free time, or when I get tired of translating and need something else to do.

1ec5:
No worries, it’s totally fine to start out with just the most common abbreviations, things like “Pq” for “parque”. There are plenty of abbreviations in English that we omit because we’re not sure a user would comprehend it correctly at a glance, and also plenty that we omit because they’re only applicable to a particular region.

PR #230, opened 2018-03-23
Standardize Spanish on informal imperative

1ec5:
The Castilian Spanish localization consistently uses the informal imperative form, while the (Latin American) Spanish localization uses a mix of formal and informal imperative forms. This change standardizes both localizations on the informal imperative form for consistency. If there’s a need for giving turn instructions more formally, that could be a separate `es-formal` localization.

Also updated remaining untranslated strings in both Spanish localizations.

## Tasklist

 - [x] Find and replace formal usage in Transifex
 - [x] Pull translations from Transifex and update test fixtures
 - [x] Add changelog entry
 - [ ] Review

/cc @danpaz @Guardiola31337 @Sealos


languages/translations/es-ES.json
```diff
@@ -417,24 +417,24 @@
...
+                "default": "Gire a la derecha",
+                "name": "Gire a la derecha en {way_name}",
+                "destination": "Gire a la derecha hacia {destination}"
             },
             "straight": {
-                "default": "En la rotonda continúa recto",
-                "name": "En la rotonda continúa recto por {way_name}",
-                "destination": "En la rotonda continúa recto hacia {destination}"
+                "default": "Continúe recto",
+                "name": "Continúa recto por {way_name}",
```
danpaz: Mixed tenses here: `Continúe` `Continúa` - was there an issue with the find and replace?

1ec5: Yes, I think so. Transifex’s find and replace is not as reliable as one would expect. 😞

1ec5: I did find-and-replace on the `es` localization but not the `es-ES` localization, which looked at a glance to already have been standardized. I’ll fix up the `es-ES` localization.

1ec5: Fixed in d1335e6dce662fd57d1563a2eea9253d4c267ee9.

1ec5:
As a sanity check, I’ve verified that both Google Maps and Apple Maps use the informal imperative form across all the Spanish dialects you can choose from on iOS.

PR #231, opened 2018-03-23
Update Dutch, Romanian, and other translations

1ec5:
Pulled translations from Transifex in every language but Spanish, which is covered by #230. Several of these languages were already up-to-date except for the recent addition of an exit number phrase. This is apparently the first time any translations have come in for that phrase, so the test script needed to also update fixtures for it.

Most of the other updates are for somewhat older changes, such as the removal of “At the roundabout” for `roundabout turn` maneuvers. The Dutch and Romanian localizations make significant gains in coverage as a result. Otherwise, there remain many gaps in translation, for which I used a three-way merge tool to avoid overwriting functional translations with English.

## Tasklist

 - [x] Pull translations from Transifex
 - [x] Modify test script to update `exit with number` phrase
 - [x] Update test fixtures
 - [ ] Add changelog entry?
 - [x] Review

## Requirements / Relations

Partially supersedes #200.

/cc @lyzidiamond @danpaz @Fabiangiger @mdomnita @ggl


ggl:
I've checked the tests for the Romanian trnaslation and the exit rotary [modifier] topic was all messed up with slight/sharp left/right and straight, so I dropped the modifiers. Also merge forward sounded odd "merge toward forward". Hopefully it fixes things buy I've yet to see another test run to make sure.

1ec5:
OK, updated in 83be412cba36a5664aa1468c7d688e9907d61c58.

> Also merge forward sounded odd "merge toward forward".

Fortunately, this particular combination should never happen. (There’s only “merge left” or “merge right”, thanks to Project-OSRM/osrm-backend#2133 and Project-OSRM/osrm-backend#2657.) So it’s OK if some of the combinations sound strange, as long as any reasonable combination sounds correct.

ggl:
Seems like it had a separate string for "merge forward" which I simply left to "Merge" and reverted the "Merge [modifier]" left/right because it did not sound quite right. Also shortened the "take the exit" statements to "take the exit" instead of "take the exit ramp".

1ec5:
Updated.

Issue #84, opened 2016-11-18
Dealing with change in translations


freenerd:
We will soon run into problems with translations getting out of sync with expected formats, since we can't get the translation in fast enough. Candidates for such change are https://github.com/Project-OSRM/osrm-text-instructions/issues/41 and https://github.com/Project-OSRM/osrm-text-instructions/issues/61.

Our gold standard is the english translation, all others follow. Just now, I think don't want to commit to guarantees regarding translations. We should do that eventually (e.g. every major release will include consistent translations).

For now we should establish fallbacks what happens when a translation is not available. I see two options:

1. Fallback to english translation
2. Fallback to not announcing the step as expected, but treat it as a `turn` instead

I guess option 2. might involve some tricky edge case handling around detecting that a translation is not there. It may also result in bad guidance. Option 1. on the other hand will certainly alienate some users. On the other hand, we can check for this case very easily during running tests and detect when a feature is missing. So I'm leaning towards option 1.

Opinions?

1ec5:
The iD project manages its localizations in Transifex. The iD project has a [script](https://github.com/openstreetmap/iD/blob/master/data/update_locales.js) that pulls in the localizations from Transifex. Transifex handles falling back to the original English string when a translation is missing, perhaps with logic for falling back to related languages; not sure. (There’s a special case for one localization to only pull in proofread translations.)

When the structure of a localizable file changes, all the affected messages get marked for retranslation, but Transifex maintains translation memory, so the translator can usually restore each message in one click. When iD releases, it releases, even if some localizations are incomplete. It would also be reasonable to establish a threshold where localizations that have less than (say) 50% coverage or are missing critical messages are automatically omitted from the built library.

/cc @bhousel

bsudekum:
@1ec5 this seems like a great idea.

PR #235, opened 2018-04-10
Arrival instruction includes an optional named waypoint

danpaz:
## Issue

Updates `compile` to accept an optional parameter `options.waypointName` to use in the arrive instruction. When the new parameter is provided, compile replaces the default arrival instruction `You have arrived at your {nth} destination` with `You have arrived at {waypoint_name}`.

For example:
```
You have arrived at your first destination
```
becomes
```
You have arrived at Somewhere
```

For the localizations, I copied the default arrival instruction and modified it manually to create the new `named` template. In most languages this was pretty straightforward to do, but for some languages I made an effort but am not 100% confident with the result. I'll highlight these in comments below for feedback from anyone with better knowledge of those languages.

## Tasklist

 - [ ] Add changelog entry
 - [ ] Review

/cc @1ec5 @lyzidiamond @bsudekum @allierowan @mcwhittemore


languages/translations/ru.json
```diff
@@ -62,49 +62,57 @@
                 "default": "Вы прибыли в {nth} пункт назначения",
                 "upcoming": "Вы прибудете в {nth} пункт назначения",
                 "short": "Вы прибыли",
-                "short-upcoming": "Вы скоро прибудете"
+                "short-upcoming": "Вы скоро прибудете",
+                "named": "Вы прибыли в {waypoint_name}"
```
yuryleb: It's better to change this text and below to _"Вы прибыли в пункт назначения, {waypoint_name}"_. Otherwise currently used form requires prepositional case applying to `{waypoint_name}` value.

And maybe it's better to publish new strings templates in Transifex first?

1ec5: The correct token to use here is probably either `{waypoint_name:date}` or `{waypoint_name:accusative}`, but I’d defer to @yuryleb. We can also fix this afterwards, once the strings are in Transifex.

languages/translations/vi.json
```diff
@@ -62,49 +62,57 @@
                 "default": "Đến nơi {nth}",
                 "upcoming": "Đến nơi {nth}",
                 "short": "Đến nơi",
-                "short-upcoming": "Đến nơi"
+                "short-upcoming": "Đến nơi",
+                "named": "Đến nơi {waypoint_name}"
```
1ec5: “Đến `{waypoint_name}`” would be better, but I can also fix this up in Transifex after the fact.

languages/translations/zh-Hans.json
```diff
@@ -62,49 +62,57 @@
                 "default": "您已经到达您的{nth}个目的地",
                 "upcoming": "您已经到达您的{nth}个目的地",
                 "short": "已到达目的地",
-                "short-upcoming": "您已经到达您的{nth}个目的地"
+                "short-upcoming": "您已经到达您的{nth}个目的地",
+                "named": "已到达目{waypoint_name}"
```
1ec5: I’m not sure if the 目 is necessary, but we can address this as part of #233 if necessary.

/cc @wingwuyf

languages/translations/ru.json
danpaz: Thanks @yuryleb, does that mean the grammar rules applied [on this line](https://github.com/Project-OSRM/osrm-text-instructions/blob/master/index.js#L254) will not take care of it for us?

yuryleb: Actually no, grammar rules work only with street names but `waypoint_name` could be any name :sad: So it's much easy just to change source strings to force nominal case for names than to extend grammar rules to handle any names.

1ec5: Good point, I didn’t really consider how the open-endedness could break the grammar rules’ assumptions. Since the client knows more about the waypoint name’s semantics, it could conceivably call `grammarize()` depending on whether it’s a street name or something else, but that wouldn’t happen by default. So I agree that putting the name in an appositive, as you suggested in https://github.com/Project-OSRM/osrm-text-instructions/pull/235#discussion_r180550563, neatly sidesteps the problem. I think we’ll need to do the same thing for at least one other language anyways.

1ec5: > And maybe it's better to publish new strings templates in Transifex first?

One of the unfortunate aspects of OSRMTI currently is that adding a new string requires immediately adding _something_ in each of the languages, whether that’s a copy of the English string or something hacked together based on preexisting translations. Transifex can only pick up new strings once the initial PR lands.

We could just insert English everywhere as a placeholder, but that would result in a temporary regression in all other languages. Regressing translation coverage isn’t necessarily a big deal, but some languages have taken months to update, even if it’s just a matter of adapting an existing string as this PR does. #84 and #155 track figuring out a better process for adding new strings.

wingwuyf:
@1ec5 目 is not necessary, 目的地 is a single Chinese vocabulary. I think it should be `"named": "已到达{waypoint_name}"`

1ec5:
9a6a98072c83f6032fc00550cb0636679293f3a6 updates the test scripts so that `UPDATE=1` also updates any `named` translations. b9d776e621db9daf466c7aa2b9d7ee12f9368b29 fixes the Simplified Chinese named waypoint arrival translations, per https://github.com/Project-OSRM/osrm-text-instructions/pull/235#issuecomment-380301060.

danpaz:
Thanks everyone! I updated the Russian translations per https://github.com/Project-OSRM/osrm-text-instructions/pull/235#discussion_r180550563. Planning to merge and cut a release with this feature shortly.

Issue #223, opened 2018-02-26
Translations donated by the European Commission


freenerd:
The European Commission has translated osrm-text-instructions to the EU member state languages and asked me to make them available to the community. Sadly the translations were for an older OSRM version and currently only only get ~77% coverage, so they are likely not ready for production. They miss: combined instructions, exits, arrival,  traffic circle.

Still I added the following translations to Transifex:

- Bulgarian `bg`
- Czech `cs`
- Greek `el`
- Estonian `et`
- Finnish (fi)
- Hungarian (hr)
- Latvian (lv)
- Maltese (mt)
- Portuguese (Portugal) (pt_PT)
- Slovak (sk) 
- Slovenian (sl) 

Languages that already had translations and I therefore did not update:

- Italian (it)
- Lithuanian (lt)
- Polish (pl)

1ec5:
Wonderful!

> Hungarian (hr)

Hungarian is hu; Croatian is hr.

> Languages that already had translations and I therefore did not update

It would be great if these translations could be made public somewhere. We can reach out to the existing localization’s translators asking if they’d like to crib anything from the European Commission’s translations.

yuryleb:
Maybe something could be copied from OSM website translations, say https://translatewiki.net/wiki/Osm:Javascripts.directions.instructions.turn_right_with_exit/hu
This also should be in publick domain.

freenerd:
> Hungarian is hu; Croatian is hr.

Ah shit. I fucked up. I had to delete the Hungarian translation, since the Croatian import overwrote everything. I contacted the translator `yzsolt` to apologise.

> It would be great if these translations could be made public somewhere.

Here are pl, it and lt: [translations.zip](https://github.com/Project-OSRM/osrm-text-instructions/files/1763163/translations.zip)

yzsolt:
No problem, I could restore most of the old translations from history, and added some new too. There are lots of duplications in the source though, translating it is pretty repetitive.

1ec5:
#229 completed the European Portuguese localization.
#239 completed the Finnish localization.
#264 completes the Slovenian localization.

PR #239, opened 2018-04-21
Add Finnish translations

akikoskinen:
Proposing Finnish translations. I continued the translations mentioned in issue #223.

This commit has issues, not all the translations are very fluent. Consider this version for review, but there's definitely room for improvement. I also post this here so that it gets visibility and perhaps some people can help in improving this.

This is the best I can do with just Transifex, but some of these aren't very good Finnish. For example the modifier `left` is translated as `vasemmall(e/a)`. That's because the modifier is used in different contexts and a different form of the word would be needed. Roughly: "to the left" is "vasemmalle", "on the left" is "vasemmalla".

This and other issues can maybe be solved with the grammar system, but I haven't yet studied how that works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to this project will be documented in this file. For change log formatting, see http://keepachangelog.com/
 
+## master
+
+- Added a Finnish localization. [#239](https://github.com/Project-OSRM/osrm-text-instructions/pull/239)
+
 ## 0.13.0 2018-04-11
 
 - Added a European Portuguese localization. [#229](https://github.com/Project-OSRM/osrm-text-instructions/pull/229)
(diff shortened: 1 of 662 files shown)


yuryleb:
Currently grammar works with `way_name` and `rotary_name` values only but why not to extend it somehow to process whole phrase or a new keyword? Anyway this could be done in a new pull request but this one should be finished with ChangeLog update :wink:

akikoskinen:
I would say either way is good. The current Finnish in this PR isn't fluent at all places but it's understandable. So I guess it's better than nothing.

Maybe it boils down to the quality ambitions of this project and the release schedule. If the quality ambition is very high and a release is approaching, then I wouldn't include this PR yet. If the approach is more that any value added is good (and can be improved later) or no release is right behind the corner, then I would say that this can go now as it is and once #240 is in, then Finnish (and other) strings can be improved.

So I think I'm throwing the ball right back at ya :smile:

Issue #251, opened 2018-06-18
Grammar rules: limits?


Penegal:
Hello, there.

Making some tests with OSRM in French, I noticed it had no support of elision, which is particularly perceptible in this language: OSRM will, for instance, say _Tourner légèrement à droite en direction de Épinal_, whereas the correct French would be _Tourner légèrement à droite en direction d’Épinal_. That would need adaptation of translated text according to rules applied on the OSM names, but would this be possible using grammar rules?

Another related question: in some cases, OSRM will say, for instance, _Tourner à gauche sur Rue de Jarménil (D 159b)_, but the correct French sentence would be _Tourner à gauche sur la rue de Jarménil (D 159b)_, with a definite article, as the way name doesn't have one, and a lower case first letter. That would greatly increase readability and user-friendliness of instructions, but would it be possible and desirable?

I'm willing to help for these, at least for the regex part, as it is my first time on this project and I'm unsure I'm able to completely write such a feature from scratch.

Regards.

yuryleb:
Yes, this looks possible and desirable. I could propose to add some (say `elision`) "grammar case" for French, add this `elision` option to all `way_name` keywords and prototype a few expressions for elisions and articles. I hope this could work.

yuryleb:
Actually your first issue requires changing whole source phrase not `way_name` value only and so this is out of grammar rules functionality :frowning:

Maybe it's possible to apply elision right in source phrase like below?
`"destination": "Tourner à droite en direction d’{destination}"`

BTW also French translation on [Transifex](https://www.transifex.com/project-osrm/osrm-text-instructions/) looks unsynchronized with current `languages/translations/fr.json` content that makes  working with new French override script to add grammar options much harder. Maybe you can fill missing translations on Transifex first or maybe @benjamintd, @patjouk, @guillaumerose could help with this?

Penegal:
I don't think it will be possible to apply elision directly in source phrases, at least not always, as it only applies before vowels and some words starting with a h. It would need a bit of regex to detect on which cases apply elision.

yuryleb:
OK, then we need for another bunch of rules to post-process whole final phrase content. Actually this is necessary also for Russian (#240) :wink:

benjamintd:
@yuryleb would it make sense to add the article (`de` or `d'`) inside the way name grammar case?

We would have something like:
`"destination": "Tourner à droite en direction {destination:article}"`
which would give: 
[...]

PR #252, opened 2018-06-19
Initial French grammar rules to insert articles to way names

yuryleb:
# Issue

Initial French grammar rules prototype to insert articles to way names (#251).

Currently only `Rue` status street name is supported:
french articles

## Tasklist

 - [ ] Add other status street names handling (?)
 - [ ] Add changelog entry
 - [x] Test with osrm-frontend
 - [ ] Review

## Requirements / Relations

New `languages/overrides/fr.js` script adds `article` option to most `way_name` keywords during translations export from Transifex, `languages/grammar/fr.json` contains regular expressions (currently one) to process street names, changed `languages.js` loads and registers that French expressions file. `test/grammar_test.js` is extended with test for French.

### DIFF
--- a/languages/grammar/fr.json
+++ b/languages/grammar/fr.json
@@ -1,0 +1,10 @@
+{
+    "meta": {
+        "regExpFlags": "gi"
+    },
+    "v5": {
+        "article": [
+            ["^ Rue ", " la rue "]
+        ]
+    }
+}
--- a/languages/translations/fr.json
+++ b/languages/translations/fr.json
@@ -118,59 +118,59 @@
         "continue": {
             "default": {
                 "default": "Tourner {modifier}",
-                "name": "Tourner {modifier} pour rester {way_name}",
+                "name": "Tourner {modifier} pour rester {way_name:article}",
                 "destination": "Tourner {modifier} en direction de {destination}",
-                "exit": "Tourner {modifier} sur {way_name}"
+                "exit": "Tourner {modifier} sur {way_name:article}"
             },
             "straight": {
                 "default": "Continuer tout droit",
-                "name": "Continuer tout droit pour rester sur {way_name}",
+                "name": "Continuer tout droit pour rester sur {way_name:article}",
                 "destination": "Continuer tout droit en direction de {destination}",
                 "distance": "Continuer tout droit sur {distance}",
-                "namedistance": "Continuer sur {way_name} sur {distance}"
+                "namedistance": "Continuer sur {way_name:article} sur {distance}"
             },
             "sharp left": {
                 "default": "Braquer à gauche",
-                "name": "Braquer à gauche pour rester sur {way_name}",
+                "name": "Braquer à gauche pour rester sur {way_name:article}",
                 "destination": "Braquer à gauche en direction de {destination}"
             },
             "sharp right": {
                 "default": "Braquer à droite",
-                "name": "Braquer à droite pour rester sur {way_name}",
+                "name": "Braquer à droite pour rester sur {way_name:article}",
                 "destination": "Braquer à droite en direction de {destination}"
             },
             "slight left": {
                 "default": "S’aligner légèrement à gauche",
-                "name": "S’aligner légèrement à gauche pour rester sur {way_name}",
+                "name": "S’aligner légèrement à gauche pour rester sur {way_name:article}",
                 "destination": "S’aligner légèrement à gauche en direction de {destination}"
             },
             "slight right": {
                 "default": "S’aligner légèrement à droite",
-                "name": "S’aligner légèrement à droite pour rester sur {way_name}",
+                "name": "S’aligner légèrement à droite pour rester sur {way_name:article}",
                 "destination": "S’aligner légèrement à droite en direction de {destination}"
             },
             "uturn": {
                 "default": "Faire demi-tour",
-                "name": "Faire demi-tour et continuer sur {way_name}",
+                "name": "Faire demi-tour et continuer sur {way_name:article}",
                 "destination": "Faire demi-tour en direction de {destination}"
             }
         },
         "depart": {
             "default": {
                 "default": "Rouler vers {direction}",
-                "name": "Rouler vers {direction} sur {way_name}",
-                "namedistance": "Rouler vers {direction} sur {way_name} sur {distance}"
+                "name": "Rouler vers {direction} sur {way_name:article}",
+                "namedistance": "Rouler vers {direction} sur {way_name:article} sur {distance}"
             }
         },
         "end of road": {
             "default": {
                 "default": "Tourner {modifier}",
-                "name": "Tourner {modifier} sur {way_name}",
+                "name": "Tourner {modifier} sur {way_name:article}",
                 "destination": "Tourner {modifier} en direction de {destination}"
             },
             "straight": {
                 "default": "Continuer tout droit",
-                "name": "Continuer tout droit sur {way_name}",
+                "name": "Continuer tout droit sur {way_name:article}",
                 "destination": "Continuer tout droit en direction de {destination}"
             },
             "uturn": {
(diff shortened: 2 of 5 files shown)


yuryleb:
The next step:
french elisions

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+    "meta": {
+        "regExpFlags": "gi"
+    },
+    "v5": {
+        "article": [
+            ["^ Acc[èe]s ", " l’accès "],
+            ["^ All[ée]e ", " l’allée "],
+            ["^ (L['’])?Autoroute ", " l’autoroute "],
+            ["^ Avenue ", " l’avenue "],
+            ["^ Boulevard ", " la boulevard "],
```
Penegal: We say « le boulevard ».

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Chemin ", " la chemin "],
```
Penegal: We say « le chemin ».

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Cours ", " la cours "],
+            ["^ Entr[ée]e ", " l’entrée "],
+            ["^ Impasse ", " l’impasse "],
+            ["^ Parvis ", " la parvis "],
```
Penegal: Same remark here.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Passage ", " la passage "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Place ", " la place "],
+            ["^ Petit[\\- ]Pont ", " la petit-pont "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Pont ", " la pont "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Promenade ", " la promenade "],
+            ["^ Quai ", " la quai "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Route ", " la route "],
+            ["^ Rue ", " la rue "],
+            ["^ Sortie ", " la sortie "],
+            ["^ Souterrain ", " la souterrain "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Square ", " la square "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Tunnel ", " la tunnel "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Voie ", " la voie "]
+        ],
+        "elision": [
+            ["^ Le ", "  du "],
+            ["^ Les ", "  des "],
+            ["^ La ", "  de "],
```
Penegal: There is no elision here: « de la » is grammaticaly correct. You only need to get rid of the first capital.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+
+            ["^ Acc[èe]s ", "  de l’accès "],
+            ["^ All[ée]e ", "  de l’allée "],
+            ["^ (L['’])?Autoroute ", "  de l’autoroute "],
+            ["^ Avenue ", "  de l’avenue "],
+            ["^ Boulevard ", "  de la boulevard "],
```
Penegal: We say « du boulevard ».

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Chemin ", "  de la chemin "],
```
Penegal: We say « du chemin ».

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Cours ", "  de la cours "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Entr[ée]e ", "  de l’entrée "],
+            ["^ Impasse ", "  de l’impasse "],
+            ["^ Parvis ", "  de la parvis "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Passage ", "  de la passage "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Place ", "  de la place "],
+            ["^ Petit[\\- ]Pont ", "  de la petit-pont "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Pont ", "  de la pont "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Promenade ", "  de la promenade "],
+            ["^ Quai ", "  de la quai "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Route ", "  de la route "],
+            ["^ Rue ", "  de la rue "],
+            ["^ Sortie ", "  de la sortie "],
+            ["^ Souterrain ", "  de la souterrain "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Square ", "  de la square "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Tunnel ", "  de la tunnel "],
```
Penegal: Idem.

languages/grammar/fr.json
```diff
@@ -0,0 +1,64 @@
...
+            ["^ Voie ", "  de la voie "],
+
+            ["^ ([AÂÀEÈÉÊËHIÎÏOÔUÙÛÜYŸÆŒ])", "  d’$1"],
```
Penegal: It seems that there is no real rule for elision before an h, so it can't be easily done apart with a list of all concerned words. I think you can remove it from this list and the other one. If you want to do it, here is a list of all h-beginning words which don't provoke an elision: https://fr.wikipedia.org/wiki/H_aspir%C3%A9

languages/grammar/fr.json
benjamintd: to be more precise, it would be "du quai"

languages/grammar/fr.json
benjamintd: "du souterrain"

languages/grammar/fr.json
benjamintd: "du square"

languages/grammar/fr.json
benjamintd: "du tunnel"

languages/grammar/fr.json
yuryleb: This list looks too big to be inserted into this humble expressions array :confused: So it's really simpler just to exclude H from there

Penegal:
@benjamintd: I would say the correct string would be `de La Rochelle`: `La` is an article which is not subject to elision, so it would have to stay, and, as it is part of the name, it should keep its capital first letter.

This would have been different if the name was, for instance, `Les Marches`, as the elision would form `des Marches`, without capital since `de` has none, but, since the `La` article is not subject to this kind of elision, it should keep its capital first letter.

Penegal:
I also notice that grammar rules should be applied on two other cases:
1. when a name is placed after `at`: for instance, the correct translation for `You have arrived at Les Marches` is `Vous êtes arrivé aux Marches`, with elision, and not `Vous êtes arrivé à Les Marches`;
2. when a name is placed after `onto`: the correct translation of `Merge left onto Avenue des Champs-Élysées` is `S’insérer à gauche sur l’avenue des Champs-Élysées`, not `S’insérer à gauche sur Avenue des Champs-Élysées`: the second case have a faulty capital and misses the required article `la`, which is elided in `l’` in this case.

languages/grammar/fr.json
yuryleb: @Penegal, does it mean " la cours" in `article` group above should be also changed to "le cours"?

yuryleb:
I need more detailed cases for replacing `à` to `aux` in `Vous êtes arrivés à {waypoint_name}` strings :wink: Or just `aux` should be used everywhere with `{waypoint_name}`?

Penegal:
@yuryleb: `à le` is to be replaced by `au`, and `à les` by `aux`. This is not related to the first letter, as with `l’`, but to the first word; it is to be applied even if `le` or `les` is in the variable instead of the OSRM string, and is case insensitive: `à Le` is also to be replaced by `au`, and `à Les` by `aux`. You can see an example in my last comment.

languages/translations/fr.json
```diff
@@ -118,367 +118,367 @@
         "continue": {
             "default": {
                 "default": "Tourner {modifier}",
-                "name": "Tourner {modifier} pour rester {way_name}",
-                "destination": "Tourner {modifier} en direction de {destination}",
-                "exit": "Tourner {modifier} sur {way_name}"
+                "name": "Tourner {modifier} pour rester {way_name:article}",
```
Penegal: The correct translation would be `Tourner {modifier} pour rester sur {way_name:article}`.

languages/translations/fr.json
```diff
@@ -118,367 +118,367 @@
...
-                "name": "Continuer tout droit sur {way_name}",
-                "destination": "Continuer tout droit en direction de {destination}"
+                "name": "Continuer tout droit sur {way_name:article}",
+                "destination": "Continuer tout droit en direction {destination:elision}"
             },
             "uturn": {
                 "default": "Faire demi-tour à la fin de la route",
-                "name": "Faire demi-tour à la fin de la route {way_name}",
-                "destination": "Faire demi-tour à la fin de la route en direction de {destination}"
+                "name": "Faire demi-tour à la fin de la route {way_name:article}",
```
Penegal: The correct translation would be `Faire demi-tour à la fin {way_name:elision}`.

Penegal: Correction: `Faire demi-tour à la fin de {way_name:article}`.

yuryleb: So `Faire demi-tour à la fin de le boulevard Saint-Germain` not `Faire demi-tour à la fin du boulevard Saint-Germain`?

yuryleb:
Example of `rotary`:
french rotary

Penegal:
Here is what I expected: the correct second instruction would be `Prendre le rond-point de la place Charles de Gaulle`. It should change according to the name of the roundabout; if it is named like `Rond-point x`, then the correct sentence would be `Prendre le rond-point x`, but, if it is another name, it should use the new elision rules: `Place x` would give `Prendre le rond-point de la place x`.

Penegal:
You should go with `Faire demi-tour à la fin {way_name:preposition}`, it's the correct sentence.

yuryleb:
Yes, I can support both but what should be in final text - `rondpoint` or `rond-point`?

And actually  I could also merge your changes from Transifex if you completed - anyway I download them every time I check how override script works :wink:

Penegal:
The main orthograph is `rond-point`; the other one is controversial and most people would see it as an error.

If you can merge the Transifex changes with the one you did, perfect! I have a backup of the first ones, if needed.

Penegal:
I noticed you added way types in your regex list; there could also be _le sentier_, which is something like a feminine form of _sente_.

yuryleb:
It seems all is done. @Penegal, @benjamintd, do you have anything to add?

Penegal:
@yuryleb: `de/du/des/le/la/l’` should not use a bold typeface unless they are part of the name/destination, as, in this case, they are technically not part of it.

yuryleb:
@Penegal, `village` has two L in expressions, it was just misprint in commit comment.

Formatting with bold typeface is part of external `formatToken()` implementation inside OSRM sample frontend (see https://github.com/Project-OSRM/osrm-frontend/blob/gh-pages/src/itinerary_builder.js#L11). Perhaps I could extend my last [formatToken() fix](https://github.com/Project-OSRM/osrm-frontend/pull/271) with your `de/du/des/le/la/l’` formatting proposal.

yuryleb:
@Penegal, so an expression to exclude appended articles/prepositions should look like `^(à )|(au )|(aux )|(d’)|(de )|(des )|(du )|(l’)|(la )|(le )|(les )` (case-sensitive to distinct appended words from existing in name)?

This expression will help to format as below:

- la **cour d’Honneur**
- de **la cour d’Honneur**
- le **rond-point de la cour d’Honneur**

Penegal:
@yuryleb: I would say the correct typography would be _de la **cour d’Honneur**_.

yuryleb:
@Penegal, please look how this is now [implemented](https://github.com/Project-OSRM/osrm-frontend/pull/271#issuecomment-400218680).

Penegal:
@yuryleb: very good work. I'm OK with this.

Penegal:
Is there a problem preventing an approving review?

languages/grammar/fr.json
```diff
@@ -0,0 +1,177 @@
...
+            ["^ Sente ", " la sente "],
+            ["^ Sentier ", " le sentier "],
+            ["^ Sortie ", " la sortie "],
+            ["^ Souterrain ", " le souterrain "],
+            ["^ Square ", " le square "],
+            ["^ Terrasse ", " la terrasse "],
+            ["^ Traverse ", " la traverse "],
+            ["^ Tunnel ", " le tunnel "],
+            ["^ Viaduc ", " le viaduc "],
+            ["^ Villa(ge)? ", " le villa$1 "],
```
benjamintd: What are we looking to match other than Village? Unfortunately `Villa` is feminine in French (la villa)

yuryleb: Thank you, I add other rules for `Villa` then

Penegal:
How often is the code of map.project-osrm.org updated? Just wanted to see it live… :grin:

1ec5:
map.project-osrm.org is powered by [OSRM-frontend](https://github.com/Project-OSRM/osrm-frontend/). That repository will need to be updated, but first we have to publish a new version of OSRMTI.

PR #255, opened 2018-07-19
v0.13.1

1ec5:
Prepared for the v0.13.1 release. Also updated the Node.js version. ([Travis consults .nvmrc](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions-using-.nvmrc) for the Node version.)

/cc @danpaz @bsudekum


.travis.yml
```diff
@@ -1,3 +1 @@
 language: node_js
-node_js:
-  - "4.4"
```
bsudekum: I imagine we will still want to specify a version of node here, maybe match that in the .nvmrc file.

1ec5: [Travis falls back to what’s in .nvmrc](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions-using-.nvmrc). Does anything other than Travis use this file?

bsudekum: Ok, that works then.

PR #256, opened 2018-07-23
American English and roundabouts

### LINKED ISSUE #188: Rename traffic circles to roundabouts
#179 replaced occurrences of “roundabout” with “traffic circle” in the English strings. I think we should consider returning to “roundabout” for the following reasons:

* “Roundabout” is the generally recognized term outside the United States, modulo some local color as seen in https://github.com/Project-OSRM/osrm-text-instructions/pull/179#issuecomment-338125356. We don’t want to have to maintain separate `en-GB`, `en-AU`, `en-NZ`, etc. localizations just for this one word (although I wouldn’t be opposed to starting more localizations for other reasons).
* In the United States, especially outside the Northeast, “roundabout” has achieved better recognition than “traffic circle” as a standard road feature. It’s difficult to come up with an 🍎↔🍏 comparison, because “roundabout” is also an unrelated adjective, but [“roundabout” as a noun](https://www.google.com/search?tbm=nws&q="roundabout+at"+location%3A"United+States") appears about 21 times as often as [“traffic circle”](https://www.google.com/search?tbm=nws&q=%22traffic+circle+at%22+location%3A%22United+States%22) in recent U.S. news articles.
* In the United States, “traffic circle” can [imply](https://www.fhwa.dot.gov/publications/research/safety/00067/000671.pdf) that traffic entering the junction has the right-of-way, which is only the case for the junctions OSRM calls “rotaries”. I would agree with leaving the `rotary` instructions with “traffic circle” for correctness’ sake.

Some arguments have been put forth in favor of “traffic circle”:

* For users who have yet to encounter a roundabout, the term “traffic circle” is more descriptive.
  * I would contend that the number of drivers who are unfamiliar with how to navigate a roundabout far exceeds the number of drivers who are unfamiliar with the concept in the first place.
* Google Maps says “traffic circle” [in the U.S.](https://www.google.com/maps/dir/39.3665131,-84.3206929/39.3651732,-84.3192072/@39.3663487,-84.3210892,17z/am=t/data=!4m6!4m5!3e0!6m3!1i0!2i0!3i1), [including for the traffic calming device](https://www.google.com/maps/dir/37.3539206,-121.8951318/841-801+N+7th+St,+San+Jose,+CA+95112/@37.3539106,-121.8969579,17z/am=t/data=!4m13!4m12!1m0!1m5!1m1!1s0x808fcc87327972a5:0x47b1bc438c9ed96e!2m2!1d-121.8958173!2d37.354802!3e0!6m3!1i0!2i0!3i0) that takes up the same space as a standard intersection. (They say “roundabout” [in the UK](https://www.google.com/maps/dir/51.5624299,-1.7730783/16-10+Shrivenham+Rd,+Swindon+SN1,+UK/@51.5626473,-1.7742531,16z/am=t/data=!4m13!4m12!1m0!1m5!1m1!1s0x4871444acb184f47:0x4494aad55da7164!2m2!1d-1.7708572!2d51.5632739!3e0!6m3!1i0!2i0!3i1) and elsewhere.)

The MUTCD, the standard for road markings in the U.S., includes the following [definitions](https://mutcd.fhwa.dot.gov/htm/2009/part1/part1a.htm#section1A13):

> 30. Circular Intersection—an intersection that has an island, generally circular in design, located in the center of the intersection where traffic passes to the right of the island. Circular intersections include roundabouts, rotaries, and traffic circles.
> <div></div>
> 
> 180. [Roundabout](https://mutcd.fhwa.dot.gov/htm/2009/part3/part3c.htm)—a circular intersection with yield control at entry, which permits a vehicle on the circulatory roadway to proceed, and with deflection of the approaching vehicle counter-clockwise around a central island.

As of the 2009 edition, which was the first edition to specify various details about roundabouts, either word may be used on signage, but “roundabout” is preferred:

plaque

/cc @mcwhittemore @danpat

1ec5:
# Issue

Added an American English localization that is programmatically derived from the international English localization. The English localization remains the default. The two localizations are identical, except for circular junction terminology. The English localization calls all circular junctions “roundabouts” – which is a change in favor of non-American usage – whereas the American English localization distinguishes between classical roundabouts as “traffic circles” and modern roundabouts as “roundabouts”. I think we could justify calling them all roundabouts in American English, but this is a slightly more conservative change that respects East Coast tradition.

This isn’t exactly a backwards-incompatible change, but clients that wish to retain some “traffic circle” usage should ensure the use of the `en-US` locale instead of relying on the `en` default.

Fixes #188.

## Tasklist

 - [x] Align to international English usage
 - [x] Add American English localization
 - [x] Add changelog entry
 - [ ] Review

/cc @danpaz @bsudekum


1ec5:
> This isn’t exactly a backwards-incompatible change, but clients that wish to retain some “traffic circle” usage should ensure the use of the `en-US` locale instead of relying on the `en` default.

@danpaz reminded me that the `language=en` default (when the parameter is omitted) is quite entrenched in the Mapbox Directions API and perhaps other clients as well. I overlooked this possibility because the Mapbox Navigation SDK automatically specifies a full locale code including a region, such as `en-US`.

One possibility would be to fold the remaining “traffic circle” usage back into the `en` locale and create localizations for each of the major English locales (including `GB`, `CA`, `AU`, `NZ`, `IN`, and `ZA`). But even if we automate those locales using a transformation script, as this PR attempts to do, that’s a lot of duplication in the final package for relatively little gain.

#155 would enable us to avoid duplicating the entire English localization and ship the package with “sparse” localizations that only override individual strings. In the meantime, we might want to consider paring back this PR to just replace “traffic circle” with “roundabout” everywhere until #155 gives us that extra flexibility.

andrewharvey:
> One possibility would be to fold the remaining “traffic circle” usage back into the en locale and create localizations for each of the major English locales (including GB, CA, AU, NZ, IN, and ZA). But even if we automate those locales using a transformation script, as this PR attempts to do, that’s a lot of duplication in the final package for relatively little gain.

Not sure if you could just create an en-gb, then symlink en-ca, en-au, etc all to en-gb until such time as they need to diverge?

1ec5:
Per chat with @danpaz, we’re going to change the wording to roundabout for all English locales, making an en-US locale unnecessary.

1ec5:
Circling back on this fix in #285.

PR #226, opened 2018-03-15
Add some German, Hebrew, Hungarian, Slovenian, Ukrainian abbreviations

1ec5:
Added abbreviation files for German, Hebrew, Hungarian, Slovenian, Ukrainian, based on partial abbreviation translations in the Mapbox Navigation SDK for iOS project on Transifex.

Thanks to @P-Zenker, @wishtrip-dev, @yzsolt, [AndrejK](https://www.transifex.com/user/profile/AndrejK/), and @Andygol for contributing these abbreviations before the abbreviation resource was moved to this repository as part of mapbox/mapbox-navigation-ios#1050. Going forward, if you need to make modifications, please open a pull request here – the translations will benefit both the Android and iOS navigation SDKs that way.

/cc @danpaz

PR #264, opened 2018-08-15
Add Slovenian localization

1ec5:
Added a complete Slovenian localization courtesy of Transifex contributor [peter.klofutar](https://www.transifex.com/user/profile/peter.klofutar/) and the European Commission (#223). A Slovenian abbreviations file is already part of OSRMTI, having been imported from the iOS navigation SDK in #226.

/cc @danpaz


1ec5:
I forgot to pull from Transifex one more time before merging: #265.

Issue #271, opened 2019-02-22
Hungarian language pull request


kiskunk:
I would like to request the hungarian language to be included in the OSRM-text-instructions.

yuryleb:
Hungarian translation on [Transifex](https://www.transifex.com/project-osrm/osrm-text-instructions/language/hu/) needs to complete 164 strings yet. Maybe you could fill them there and check existing strings for correctness?

kiskunk:
> Hungarian translation on [Transifex](https://www.transifex.com/project-osrm/osrm-text-instructions/language/hu/) needs to complete 164 strings yet. Maybe you could fill them there and check existing strings for correctness?

I will, but how can I check for correctness? I mean if I think some is correct, how can I mark it correct?

yuryleb:
Just don't touch such correct string(s) or press Approve or something like this :wink:

kiskunk:
> Hungarian translation on [Transifex](https://www.transifex.com/project-osrm/osrm-text-instructions/language/hu/) needs to complete 164 strings yet. Maybe you could fill them there and check existing strings for correctness?

I checked the existing ones and I filled the missing ones. I do not have review rights on Transifex though, but I would say the language strings are now pullable. :)

yuryleb:
I tried to merge your translation but found many missing(?) whitespaces after `{way_name}`, `{waypoint_name}`, `{distance}` and other macros like below:
```json
        "continue": {
            "default": {
                "default": "Forduljon {modifier}",
                "name": "Forduljon {modifier} és maradjon a(z) {way_name}szakaszon",
                "destination": "Forduljon {modifier} a(z) {destination} felé",
                "exit": "Forduljon be {modifier} a(z) {way_name} irányába"
            },
```
Could you please check this more on Transifex?

And what `a(z)` here mean? Alternate preposition/ending depending on used macro value? Maybe Hungarian translation also needs for [grammar support](https://github.com/Project-OSRM/osrm-text-instructions/blob/master/languages/grammar/README.md)?

kiskunk:
[...]

PR #274, opened 2019-02-28
Add Hungarian localization

yuryleb:
# Issue

Initial Hungarian translation contributed by @kiskunk initally discussed in #271.
Thanks a lot!

## Tasklist

 - [x] Review
 - [x] Add changelog entry
 - [x] Test with osrm-frontend (?)


yuryleb:
No problem, that's the review goal is 😉

kiskunk:
> No problem, that's the review goal is 😉

Ok, will I get a new notification if the new review is available, or I have to check back regularly? Thanks.

yuryleb:
I'll add comment here and then you'll get a notification. Did you already update translation on Transifex?

kiskunk:
> I'll add comment here and then you'll get a notification. Did you already update translation on Transifex?

Yes, I did it before commenting on the review here.

yuryleb:
IMHO missing whitespace after `{modifier}` in new `name` stuff:
```json
        "merge": {
            "default": {
                "default": "Soroljon be {modifier}",
                "name": "Soroljon be {modifier}a(z) {way_name} felé",
                "destination": "Soroljon be {modifier} ehhez: {destination}"
            },
```

kiskunk:
> IMHO missing whitespace after `{modifier}` in new `name` stuff:
> 
> ```json
>         "merge": {
>             "default": {
>                 "default": "Soroljon be {modifier}",
>                 "name": "Soroljon be {modifier}a(z) {way_name} felé",
>                 "destination": "Soroljon be {modifier} ehhez: {destination}"
>             },
> ```

Fixed it, tho it's really hard to keep track since almost every value exist 6 times.

yuryleb:
Actually no, only `destination` was changed but `name` still need for whitespace:
```json
        "merge": {
            "default": {
                "default": "Soroljon be {modifier}",
                "name": "Soroljon be {modifier}a(z) {way_name} felé",
                "destination": "Soroljon be {modifier} a(z) {destination} irányába"
            },
```

kiskunk:
I only saw one, fixed, then searched for }a and found nothing. Is it possible to remove the ultimate redundancy somehow so every string would be there only 1 times?

yuryleb:
I pushed your updated translation, please review result again.

IMHO there are extra whitespaces in `name and ref` now :wink:
```json
        "phrase": {
            "two linked by distance": "{instruction_one} , majd {distance} múlva, {instruction_two}",
            "two linked": "{instruction_one} , majd {instruction_two}",
            "one in distance": "{distance} múlva {instruction_one}",
            "name and ref": "{name} ( {ref} )",
            "exit with number": "{exit} kijárat"
        },
```

kiskunk:
Sorry, i wanted to make sure there is always whitespace, tomorrow I will check them, so to be clear there can be braces and commas after them ?

yuryleb:
I suppose there should be just `"name and ref": "{name} ({ref})",`

test/fixtures/v5/other/way_name_ref_name.json
```diff
@@ -18,6 +18,7 @@
         "fi": "Käänny vasempaan tielle Way Name (Ref1)",
         "fr": "Tourner à gauche sur Way Name (Ref1)",
         "he": "פנה שמאלה לWay Name (Ref1)",
+        "hu": "Forduljon be balra a(z) Way Name ( Ref1 ) irányába",
```
yuryleb: Here is extra whitespaces sample

languages/translations/hu.json
```diff
@@ -119,47 +119,47 @@
...
             }
         },
         "depart": {
             "default": {
-                "default": "Hajtson {direction} felé",
-                "name": "Hajtson {direction} felé a(z) {way_name} szakaszon",
-                "namedistance": "Hajtson {direction} felé a(z) {way_name} , {distance} távolságig"
+                "default": "Hajtson a(z) {direction} irányába",
+                "name": "Hajtson a(z) {direction} irányába a(z) {way_name} szakaszon",
+                "namedistance": "Hajtson a(z) {direction} irányába a(z) {way_name} felé {distance}  távolságig"
```
yuryleb: Extra whitespace after `{distance}` IMHO

languages/translations/hu.json
```diff
@@ -418,8 +418,8 @@
                 },
                 "default": {
                     "default": "Hajtson be a körforgalomba",
-                    "name": "Hajtson a körforgalomba, majd hagyja el a körforgalmat a(z) {way_name} felé",
-                    "destination": "Hajtson a körforgalomba, majd hagyja el a körforgalmat a(z) {destination} irányába"
+                    "name": "Hajtson a körforgalomba, majd hagyja el a körforgalmat a(z) {way_name}  felé",
+                    "destination": "Hajtson a körforgalomba, majd hagyja el a körforgalmat a(z) {destination}  irányába"
```
yuryleb: And here after `{way_name}` and `{destination}` too

yuryleb:
Next commit pushed. But now there are extra whitespaces 😞

kiskunk:
Transifex really plays out its cards... oh well, tried to fix them all.

yuryleb:
Committed last fix. Translation looks good for me now 👍 

BTW maybe you could also add translation to the frontend? It's necessary just add new hu.js into [i18n](https://github.com/Project-OSRM/osrm-frontend/tree/gh-pages/i18n) folder with similar content:
```
module.exports = {
  name: 'English',
  key: 'en',
  'Open in Debug Map': 'Open in Debug Map',
  'Open in Mapillary': 'Open in Mapillary',
  'GPX': 'Export GPX file',
  'Open in editor': 'Open in editor',
  'Open in JOSM': 'Open in JOSM',
  'Select language': 'Select language',
  'Start - press enter to drop marker': 'Start - press enter to drop marker',
  'End - press enter to drop marker': 'End - press enter to drop marker',
  'Via point - press enter to drop marker': 'Via point - press enter to drop marker'
};
```
and register it in [src/localization.js](https://github.com/Project-OSRM/osrm-frontend/blob/gh-pages/src/localization.js)?

You could do this in your own new pull request or just send/type translation for me.
Then I could test your localization for osrm-text-instructions together with frontend.

kiskunk:
> Translation is also checked with frontend:
> osrm-hungarian
> 
> So if all is ok, maybe this could be merged?
> 
> Also IMHO new version of whole package should be released already, maybe maintainers could do this finally?
> 
> /cc @1ec5

It looks and reads correct.

yuryleb:
Maybe @danpaz could merge this PR?

languages/translations/hu.json
```diff
@@ -0,0 +1,493 @@
...
+                "xo": "Tartson jobbra",
+                "ox": "Tartson balra",
+                "xox": "Maradjon középen",
+                "oxo": "Tartson balra vagy jobbra"
+            }
+        },
+        "modes": {
+            "ferry": {
+                "default": "Hajtson fel a kompra",
+                "name": "Hajtson fel a kompra a(z) {way_name} felé",
```
1ec5: According to [Wiktionary](https://en.wiktionary.org/wiki/az#Article), it would be “a” or “az” depending on the first letter of `way_name`. `a(z)` likely won’t sound right when spoken by a text-to-speech engine. We should add a simple grammar rule that changes it to “a” or “az” based on `way_name`.

yuryleb: Yes, and @kiskunk and I agreed to try to do this in separate PR. But I could try to include this into already prepared. I just worry that PR may become too huge to be reviewed completely.

1ec5: @kiskunk, would these regular expressions correctly match all the vowels and consonants that could appear at the beginning of a name?

```json
"article": [
    ["^ [aáäåâeéëiíïoóőöuúűüyÿ]", " az "],
    ["^ [^aáäåâeéëiíïoóőöuúűüyÿ]", " a "]
]
```

There would also need to be an override file similar to the French one:

https://github.com/Project-OSRM/osrm-text-instructions/blob/ed1a4a97647e6bb18eea4bcc8fa098c7d9f35f20/languages/overrides/fr.js#L3-L9

kiskunk: > @kiskunk, would these regular expressions correctly match all the vowels and consonants that could appear at the beginning of a name?
> 
> ```json
> "article": [
>     ["^ [aáäåâeéëiíïoóőöuúűüyÿ]", " az "],
>     ["^ [^aáäåâeéëiíïoóőöuúűüyÿ]", " a "]
> ]
> ```
> 
> There would also need to be an override file similar to the French one:
> 
> [osrm-text-instructions/languages/overrides/fr.js](https://github.com/Project-OSRM/osrm-text-instructions/blob/ed1a4a97647e6bb18eea4bcc8fa098c7d9f35f20/languages/overrides/fr.js#L3-L9)
> 
> Lines 3 to 9 in [ed1a4a9](/Project-OSRM/osrm-text-instructions/commit/ed1a4a97647e6bb18eea4bcc8fa098c7d9f35f20)
> 
>  var replaces = [ 
>      [' (de )?\{destination\}', ' {destination:preposition}'], // eslint-disable-line no-useless-escape 
>      [' (le rond-point )?\{rotary_name\}', ' {rotary_name:rotary}'], // eslint-disable-line no-useless-escape 
>      [' fin (de )?(la route )?\{way_name\}', ' fin {way_name:preposition}'], // eslint-disable-line no-useless-escape 
>      [' \{way_name\}', ' {way_name:article}'], // eslint-disable-line no-useless-escape 
>      [' (à )?\{waypoint_name\}', ' {waypoint_name:arrival}'] // eslint-disable-line no-useless-escape 
>  ];

It would be good for most of the time, but not always, if there is any way to determine that a way_name/destination is a highway or not, it could be helpful, without that the globally accepted "a(z)" format is ok.

kiskunk:
Also I already had to write my own logic for correcting a(z) including highway names, since my project needed to proceed further. (Its on android, not js and not regex based)

languages/translations/hu.json
yuryleb: So it seems we need add rules for Hungarian grammatical cases where we will recognize status street names and also change `a/az` articles appropriately. We can treat all unrecognized as highway refs and always use `az` there (but all foreign street names will be also here 😢) or add special rules to match "typical" highway refs (capital letter with digits or digits only, etc).

Currently I know 11 Hungarian status street names, see https://github.com/yuryleb/garmin-russian-tts-voices/blob/121a9bfe49eba702d5f244cd34ec8b0b7201aabb/src/Pycckuu__Milena%202.10/RP_RSET.TXT#L482-L493
Maybe you could extend this list more?

Also what grammatical cases could be applied for Hungarian messages? For example Russian localization supports 4 cases (accusative, dative, genitive and prepositional).

1ec5: Wait, so whether the instruction uses “a” or “az” depends on whether it’s a motorway? Or does it depend on some other aspect of the road related to motorways, like whether there’s a `ref` instead of a `name`?

1ec5:
> Its on android, not js and not regex based

If you’re using the Mapbox Navigation SDK for Android, note that improvements to OSRM Text Instructions are incorporated into the Mapbox Directions API and therefore the navigation SDK as well. Ideally the navigation SDK would take care of details like this so you don’t have to in your application code.

kiskunk:
An example of special case dor a az: usually highways in hungary starts with an M and then a number and the pronounciation is "emm" so thats why it needs an az before it in this special case.

yuryleb:
> usually highways in hungary starts with an M and then a number and the pronounciation is "emm"

Makes sense 😉 

@kiskunk, could you please also answer for my questions [above](https://github.com/Project-OSRM/osrm-text-instructions/pull/274#discussion_r267050801)? And also should Hungarian street names be changed together with status names for a grammatical case similar to [Russian streets](https://github.com/Project-OSRM/osrm-text-instructions/tree/master/languages/grammar#example)?

kiskunk:
> > usually highways in hungary starts with an M and then a number and the pronounciation is "emm"
> 
> Makes sense 😉
> 
> @kiskunk, could you please also answer for my questions [above](https://github.com/Project-OSRM/osrm-text-instructions/pull/274#discussion_r267050801)? And also should Hungarian street names be changed together with status names for a grammatical case similar to [Russian streets](https://github.com/Project-OSRM/osrm-text-instructions/tree/master/languages/grammar#example)?

So here is a complete list of possible street ending names (i mean like in Boulevard Street where "Street" is the ending) it includes those 11 too from your link.

WayNameEnding[] wayNameEndings = {
                new WayNameEnding(" árok", "árkon", "árokra"),
                new WayNameEnding(" átjáró", "átjárón", "átjáróra"),
                new WayNameEnding(" dűlő", "dűlőn", "dűlőre"),
                new WayNameEnding(" dűlőút", "dűlőúton", "dűlőútra"),
                new WayNameEnding(" erdősor", "erdősoron", "erdősorra"),
                new WayNameEnding(" fasor", "fasoron", "fasorra"),
                new WayNameEnding(" forduló", "fordulón", "fordulóra"),
                new WayNameEnding(" gát", "gáton", "gátra"),
                new WayNameEnding(" határsor", "határsoron", "határsorra"),
                new WayNameEnding(" határút", "határúton", "határútra"),
                new WayNameEnding(" kapu", "kapun", "kapura"),
                new WayNameEnding(" körönd", "köröndön", "köröndre"),
                new WayNameEnding(" körtér", "körtéren", "körtérre"),
                new WayNameEnding(" körút", "körúton", "körútra"),
                new WayNameEnding(" köz", "közön", "közre"),
                new WayNameEnding(" lakótelep", "lakótelepen", "lakótelepre"),
                new WayNameEnding(" lejáró", "lejárón", "lejáróra"),
                new WayNameEnding(" lejtő", "lejtőn", "lejtőre"),
                new WayNameEnding(" lépcső", "lépcsőn", "lépcsőre"),
                new WayNameEnding(" liget", "ligeten", "ligetre"),
                new WayNameEnding(" mélyút", "mélyúton", "mélyútra"),
                new WayNameEnding(" orom", "ormon", "oromra"),
                new WayNameEnding(" ösvény", "ösvényen", "ösvényre"),
                new WayNameEnding(" park", "parkon", "parkra"),
                new WayNameEnding(" part", "parton", "partra"),
                new WayNameEnding(" pincesor", "pincesoron", "pincesorra"),
                new WayNameEnding(" rakpart", "rakparton", "rakpartra"),
                new WayNameEnding(" sétány", "sétányon", "sétányra"),
                new WayNameEnding(" sikátor", "sikátoron", "sikátorra"),
                new WayNameEnding(" sor", "soron", "sorra"),
                new WayNameEnding(" sugárút", "sugárúton", "sugárútra"),
                new WayNameEnding(" tér", "téren", "térre"),
                new WayNameEnding(" udvar", "udvaron", "udvarra"),
                new WayNameEnding(" út", "úton", "útra"),
                new WayNameEnding(" utca", "utcán", "utcára"),
                new WayNameEnding(" üdülőpart", "üdülőparton", "üdülőpartra"),
                new WayNameEnding(" lánchíd", "lánchídon", "lánchídra"),
                new WayNameEnding(" híd", "hídon", "hídra")};

Grammatical cases as far as i could determine it would require only 2 of ours, but i do not know the name of those cases.
the first is the root name, the second is how it changes when you would say "on Boulevard Street", the 3rd is how it would change when you would say "onto Boulevard Street"
its a much harder implementation (if even possible) because its based on vowel harmony for 90%, so its not even for 100%... what i put in the translation on transifex solves this issue by using an additional word like if you would say "turn left onto Boulevard Street section"

I think with using the grammar file for standard street names to balance out a/az AND making a rule for M0, M1 ... M85 ... etc would cover almost all except the non-hungarian street names (there are only a few)

For the grammatical cases and the M0..M1 highway names its another dimension of difficulty (at least when the Mapbox SDK tries to read it out loud), because of how it would say the numbers, the numbers also would require a different suffix. For example M0 would be read like "emm nulla" but since we want to use it in context and a sentence, it would be "emm nullás", and thats not the end because it would still need the "section" word after it (which i think its already there in transifex because its a general translation) or to be better for the ears, the logic from the previous grammar cases, making it into "emm nulláson" or "emm nullásra", to write this logic it requires to have the numbers 0-20,30,40,50,60,70,80,90,100,200,300.. etc to be defined too with that 2 grammar case, maybe its enough to stay below 1000, i don't think that would come up in any hungarian highway names.

1ec5:
> For example M0 would be read like "emm nulla" but since we want to use it in context and a sentence, it would be "emm nullás", and thats not the end because it would still need the "section" word after it (which i think its already there in transifex because its a general translation) or to be better for the ears, the logic from the previous grammar cases, making it into "emm nulláson" or "emm nullásra", to write this logic it requires to have the numbers 0-20,30,40,50,60,70,80,90,100,200,300.. etc to be defined too with that 2 grammar case, maybe its enough to stay below 1000, i don't think that would come up in any hungarian highway names.

An SSML-aware text-to-speech engine might already handle optimizing numbers or alphanumeric words for speech. (The Mapbox Directions API wraps `{way_name}` etc. in `<speak-as interpret-as="address">` for that reason.) I don’t think OSRM Text Instructions is in a great position to implement number formatting for speech all by itself. That would be a valid feature request against the navigation SDK for when it falls back to the Android TTS engine (since the Mapbox Voice API doesn’t support Hungarian).

> what i put in the translation on transifex solves this issue by using an additional word like if you would say "turn left onto Boulevard Street section"

That’s reasonable. The only reason I brought up the grammar file was because of “a(z)”, which a text-to-speech engine is unlikely to pronounce correctly under any circumstance, unfortunately. That said, this PR doesn’t have to be the end of Hungarian localization work. If you think optimizing articles and numbers would be enough of a rabbit hole, we can merge this PR and would welcome any additional contributions regarding grammar.

yuryleb:
If we're talking about articles change only, it's not too hard to implement here 😉 

BTW how `a/az` should be changed before numbers in phrases like "Megérkezett a(z) 1. célponthoz, a bal oldalon" - I suppose to `a` everywhere?

But adding 2 grammatical cases support will be not too hard too if we need to change street status names. @kiskunk, am I understand properly that "turn right onto the street" instruction with `Dohány utca` street name should look like:

> Forduljon be jobbra a Dohány utcára felé

And we don't need to change the `Dohány` name?

kiskunk:
I think a(z) is ok for now. As for jumbers it would be "megérkezett az 1. célponthoz" but "megérkezett a 2. célponthoz" so its atain an another level of complication of what letter the spoken number would start with.

And for the street name it would be: Forduljon jobbra a Dohány utcára.  without the "felé" and yes Dohány would not change

yuryleb:
Ok, please look at first Hungarian grammar rules to handle articles only yet:
osrm-hu-articles

Grammar could be extended further to apply grammatical cases to Hungarian street names - I suppose this could be based on analyzing of following _magic_ words after `{way_name}` & `{rotary_name}` variables:

- felé
- irányába
- szakaszon

@kiskunk, you told about 2 grammatical cases but here are 3 variants - does some of them use the same grammatical case as another (if I'm right in my presumption at all)?

And BTW you seem to forget to include `alagút` (tunnel) status name into your wayNameEndings list.

kiskunk:
The image looks correct.


My list was from Wikipedia, so i will have to try to update that too with alagút. thanks.

Felé and Irányába are interchange-able, they both represent one of the 2 cases, szakaszon represents the other. I just added the other 2 for a little bit of word diversity, I used felé for way_names, and irányába for destination.

yuryleb:
One more question, @kiskunk: how to proceed foreign names - I suppose it's better to keep these felé/irányába/szakaszon words on their original place in instructions?

kiskunk:
yes, thats the best way i can imagine

languages/translations/hu.json
```diff
@@ -0,0 +1,493 @@
...
+            },
+            "right": {
+                "default": "Forduljon jobbra",
+                "name": "Forduljon be jobbra {way_name:article} felé",
+                "destination": "Forduljon jobbra {destination:article} irányába"
+            },
+            "straight": {
+                "default": "Hajtson egyenesen",
+                "name": "Hajtson egyenesen {way_name:article} felé",
+                "destination": "Hajtson egyenesen {destination} irányba"
```
yuryleb: @kiskunk, maybe there should be `irányába` not `irányba`? And article `a(z)` is missing here?

languages/translations/hu.json
```diff
@@ -0,0 +1,493 @@
...
+        "end of road": {
+            "default": {
+                "default": "Forduljon {modifier}",
+                "name": "Forduljon be {modifier} {way_name:article} irányába",
+                "destination": "Forduljon {modifier} {destination:article} felé"
+            },
+            "straight": {
+                "default": "Hajtson tovább egyenesen",
+                "name": "Hajtson tovább egyenesen {way_name:article} felé",
+                "destination": "Hajtson tovább egyenesen {destination} irányba"
```
yuryleb: @kiskunk, maybe there should be `irányába` not `irányba`? And article `a(z)` is missing here?

languages/translations/hu.json
```diff
@@ -0,0 +1,493 @@
...
+            },
+            "uturn": {
+                "default": "Forduljon vissza az út végén",
+                "name": "Forduljon vissza {way_name:article} végén",
+                "destination": "Forduljon vissza {destination} irányban az út végén"
```
yuryleb: @kiskunk, maybe article `a(z)` is missing here?

kiskunk:
Yes, I corrected the irányába, and added the missing a(z)

yuryleb:
OK, now I added support of Sublative and Superessive grammatical cases (yes, they're called so :wink:), please review how this looks on Hungarian streets:

And on foreign ones:

kiskunk:
Looks good to me, great job, thanks.

Issue #276, opened 2019-03-16
Request to add Japanese Localization from Transifex


joshandrews:
We have added Japanese localization on Transifex and are requesting that it gets merged in.

yuryleb:
Thanks for contribution!

Just one question: most variables have no whitespaces between translated text but there are several cases with them (see `phrase` block):
```
        "modes": {
            "ferry": {
                "default": "フェリーに乗る",
                "name": "{way_name}のフェリーに乗る",
                "destination": "{destination}に向けてフェリーに乗る"
            }
        },
        "phrase": {
            "two linked by distance": "{instruction_one} その先 {distance}で{instruction_two}",
            "two linked": "{instruction_one} その先 {instruction_two}",
            "one in distance": "{distance}先、{instruction_one}",
            "name and ref": "{name}({ref})",
            "exit with number": "{exit}出口です"
        },
        "arrive": {
            "default": {
                "default": "{nth}目の目的地に到着しました",
                "upcoming": "まもなく{nth}目の目的地に到着します",
                "short": "目的地に到着しました",
                "short-upcoming": "まもなく目的地に到着します",
                "named": "{waypoint_name}に到着しました"
            },
```
Is it a mistake/misprint or not?

yuryleb:
Also Chinese translation uses special bracket chars with intervals included:
```
        "phrase": {
            "two linked by distance": "{instruction_one}，{distance}后{instruction_two}",
            "two linked": "{instruction_one}，随后{instruction_two}",
            "one in distance": "{distance}后{instruction_one}",
            "name and ref": "{name}（{ref}）",
            "exit with number": "出口{exit}"
        },
``` 
Maybe Japanese one should use the same/similar brackets too?

joshandrews:
Thanks for the look through @yuryleb !

1. For locations where white spaces exist the idea was to separate the two instructions instructions. Before `その先` is first instruction, after `その先` is second instruction. This is considered normal in Japanese.

2. Both of these characters, `（` and `(`, can be used in Japanese, however in this case, `"name and ref": "{name}({ref})",` it makes more sense in Japanese to not include the space.

Let me know if you have any additional questions.

yuryleb:
Ok, please review new PR with your translation from Transifex.

BTW maybe you could also provide localization for OSRM frontend with a similar [way](https://github.com/Project-OSRM/osrm-text-instructions/pull/274#issuecomment-469984038)?

[...]

PR #277, opened 2019-03-19
Add Japanese localization

yuryleb:
# Issue

Initial Japanese translation contributed by @joshandrews (#276).
Thanks a lot!

## Tasklist

 - [x] Add changelog entry
 - [x] Test with osrm-frontend (?)
 - [x] Review


yuryleb:
Unfortunately I could not check frontend in Japan (router.project-osrm.org always returned Route error for any requests inside Japan cities somehow), please look at Japanese instructions over US map:
Japanese-osrmti

languages/translations/ja.json
```diff
@@ -0,0 +1,493 @@
+{
+    "meta": {
+        "capitalizeFirstLetter": false
+    },
+    "v5": {
+        "constants": {
+            "ordinalize": {
+                "1": "1 番",
```
friedbunny: I’m not sure about the project-specific requirements here, but in general there would be no space between a number and a counting word (like 番).

yuryleb: I think so. @joshandrews, could you please fix this in Transifex?

joshandrews:
Thanks for the feedback everyone. A couple updates:

1. We're updating some of the Transifex translations based on this feedback and from viewing in the frontend.  **I'll update here when we've finished that**.  
2. We've created a PR to translate osrm-frontend found [here](https://github.com/Project-OSRM/osrm-frontend/pull/299)

joshandrews:
We have updated the translations.  @yuryleb can you pull in most recent from Transifex?

languages/translations/ja.json
```diff
@@ -0,0 +1,493 @@
+{
+    "meta": {
+        "capitalizeFirstLetter": false
+    },
+    "v5": {
+        "constants": {
+            "ordinalize": {
+                "1": "1 つ目の",
```
yuryleb: There still are whitespaces after digits, is it ok?

joshandrews: Yes this is intended.  This structure should work best based on the way sentences are formed.

joshandrews:
@yuryleb Let me know if there is anything I can do to help get this past the finish line.

yuryleb:
We did all we could, the PR could merged by maintainers only

1ec5:
Thanks for all your help @joshandrews @yuryleb!

PR #278, opened 2019-03-22
Russian arrive instructions update

yuryleb:
# Issue

Just a Russian arrive instructions update with more appropriate verbs term.

## Tasklist

 - [x] Add changelog entry
 - [x] Review


languages/translations/ru.json
```diff
@@ -60,58 +60,58 @@
...
                 "named": "Вы прибыли в пункт назначения, {waypoint_name}"
             },
             "left": {
                 "default": "Вы прибыли в {nth} пункт назначения, он находится слева",
-                "upcoming": "Вы прибудете в {nth} пункт назначения, он будет слева",
-                "short": "Вы прибыли",
-                "short-upcoming": "Вы скоро прибудете",
+                "upcoming": "Вы прибываете в {nth} пункт назначения, он будет слева",
+                "short": "Вы прибыли, слева",
+                "short-upcoming": "Вы скоро прибудете, слева",
```
1ec5: @yuryleb, the `short` and `short-upcoming` messages are intended to omit information about the side of the street, because they’re meant to be announced as part of another instruction before the user gets to the destination. For example, a navigation application might combine this message like “In 500 feet, turn left at Main Street, then you will arrive”.

yuryleb: Thanks for clarification, fixed.

PR #279, opened 2019-03-27
Improve article matching in French instructions

### LINKED ISSUE #275: Extra `à` in french waypoint_name arrival instruction
I noticed a typo in the french localization of the `arrive.right.named` maneuver instruction:

https://github.com/Project-OSRM/osrm-text-instructions/blob/9de8ca2e5c9a6071b704bd302fca9a293de73365/languages/translations/fr.json#L80

There appears to be an extra `à` and whitespace. These should be removed, and the [fixtures](https://github.com/Project-OSRM/osrm-text-instructions/blob/9de8ca2e5c9a6071b704bd302fca9a293de73365/test/fixtures/v5/arrive_waypoint_name/right.json#L18) should be regenerated.

yuryleb:
# Issue

Match articles in French instructions with any number of whitespaces between article and `{waypoint_name}` to fix #275 issue.

## Tasklist

 - [x] Add changelog entry
 - [ ] Review

@1ec5, should I update change log for this 1 char change in French override script?


yuryleb:
Change log was also updated.

PR #283, opened 2019-04-02
Update Portuguese localizations

1ec5:
Synchronized the Portuguese localizations with Transifex. Thanks to @xendez and @jppcel for the updates.

/cc @danpaz


languages/translations/pt-BR.json
```diff
@@ -112,7 +112,7 @@
                 "upcoming": "Você vai chegar ao seu {nth} destino, em frente",
                 "short": "Você chegou",
                 "short-upcoming": "Você vai chegar",
-                "named": "You have arrived at {waypoint_name}, straight ahead"
+                "named": "Você chegou a (ao) {waypoint_name}, logo em frente"
```
1ec5: The phrase “a (ao)” won’t sound good when read aloud by a text-to-speech engine. Are both prepositions just as likely to occur with a waypoint name? Is there a way we can detect which is more appropriate by matching the waypoint name against regular expressions? If so, we can add [grammar rules](https://github.com/Project-OSRM/osrm-text-instructions/tree/master/languages/grammar/) for Brazilian Portuguese.

yuryleb: Yes, adding an article in proper form is not a hard task, please look at [Hungarian rules block](https://github.com/Project-OSRM/osrm-text-instructions/blob/master/languages/grammar/hu.json#L6-L13) for this.

1ec5: Right, I just don’t know what which specific rules we would use. _Ao_ is a contraction meaning “of the”, so I guess it would be similar to the French rules for _l’_.

jppcel: @1ec5 the "ao" is used when the word is in male gender and "a" when is in female gender.
An example:
You have arrived at **Frei** Caneca Shopping > Você chegou **ao Shopping** Frei Caneca
You have arrived at **Escola** Municipal Machado de Assis > Você chegou **a Escola** Municipal Machado de Assis.
We translated in this form to work in this two cases, but of course is better if use only the correct for that case.

1ec5: The destination name could be an arbitrary string, so unfortunately OSRMTI doesn’t automatically know the gender when forming instructions. In some other languages, we’ve created rules that try to detect the gender based on spelling patterns. For example, a regular expression could insert “ao” when the first word in the destination name ends in -o or “a” when it ends in -a or -ção. But it would probably get quite complex due to irregular nouns, determiners, etc.

languages/translations/pt-BR.json
```diff
@@ -112,7 +112,7 @@
                 "upcoming": "Você vai chegar ao seu {nth} destino, em frente",
                 "short": "Você chegou",
                 "short-upcoming": "Você vai chegar",
-                "named": "Você chegou a (ao) {waypoint_name}, logo em frente"
+                "named": "Você chegou a {waypoint_name}, logo em frente"
```
1ec5: I backed out this change in favor of a grammar system that can automatically add a (mostly) appropriate article contraction: #294. Please have a look @jppcel @yuryleb.

1ec5:
This PR is ready for a final review, now that the grammatical question has been split off as #294.

languages/translations/pt-BR.json
```diff
@@ -132,32 +132,27 @@
             "sharp left": {
                 "default": "Faça uma curva fechada a esquerda",
                 "name": "Faça uma curva fechada a esquerda para manter-se na {way_name}",
-                "destination": "Faça uma curva fechada a esquerda sentido {destination}",
-                "junction_name": "Faça uma curva fechada a esquerda em {junction_name}"
```
danpaz: Did you mean to remove the junction_name phrases?

1ec5: Oof, that’s a bad merge that crept in when rebasing.

1ec5: Or rather, it probably got overridden because I ran `npm run transifex` before Transifex had synchronized with the changes in #291.

danpaz:
Ah and reminder to update the changelog 📝

1ec5:
Whoops, forgot to interactively commit those changes. I’ve limited the changes to just the changes in Transifex. I added a blurb to the changelog since the affected strings are fairly prominent.

PR #284, opened 2019-07-19
Add Yoruba localization

1ec5:
Thanks to @toblinkz for contributing a complete Yoruba localization in Transifex!

@toblinkz, can you double-check that the generated sentences make sense? Thanks!

/ref mapbox/mapbox-navigation-android#2000
/cc @danpaz @Guardiola31337


test/fixtures/v5/continue/left_destination.json
```diff
@@ -37,6 +37,7 @@
         "tr": "Destination 1 istikametinde sol yöne dön",
         "uk": "Поверніть ліворуч у напрямку Destination 1",
         "vi": "Quẹo trái về Destination 1",
+        "yo": "Tan-anapa osisi ọnaDestination 1",
```
1ec5: Should the destination name be tacked onto the end of “ọna” like this, or should there be a space in between?

toblinkz:
Good afternoon, please how can I can a make a third Party TTS engine in my
application that would be responsible for yoruba pronounciation?

test/fixtures/v5/continue/left_destination.json
toblinkz: there is supposed to be a space between

yuryleb: And this should be fixed right in source translation strings on Transifex :wink:

1ec5: I fixed the whitespace issues in Transifex and this PR.

1ec5:
> Good afternoon, please how can I can a make a third Party TTS engine in my application that would be responsible for yoruba pronounciation?

I assume you’re referring to using a third-party TTS engine with the Mapbox Navigation SDK, per https://github.com/mapbox/mapbox-navigation-android/issues/2000#issuecomment-513310745. [Here’s a contrived example for iOS](https://docs.mapbox.com/ios/navigation/examples/custom-voice-controller/). The key is to subclass MapboxVoiceController and override `speak(_:)` to invoke the third-party TTS engine. If you need a corresponding hook for Android, I’d suggest following up in the mapbox-navigation-android repository.

PR #290, opened 2019-11-08
Update Japanese translations and add named intersections

danpaz:
# Issue

Updates ramp and exit instructions for the Japanese localization, courtesy of @tichimura. These changes were based on feedback from @tsuz who noted the "rampu" phrases sounded unnatural. 

@tsuz @YoshiYazawa for review. 

## Tasklist

 - [x] Add changelog entry
 - [ ] Review

## Requirements / Relations

### DIFF
--- a/languages/translations/ja.json
+++ b/languages/translations/ja.json
@@ -299,53 +299,53 @@
         },
         "off ramp": {
             "default": {
-                "default": "出口です",
-                "name": "{way_name}方面、出口です",
-                "destination": "{destination}方面、出口です",
-                "exit": "{exit} 出口です",
-                "exit_destination": "{destination} 方面、{exit}　出口です"
+                "default": "出口に向かう",
+                "name": "{way_name}方面、出口に向かう",
+                "destination": "{destination}方面、出口に向かう",
+                "exit": "{exit} 出口に向かう",
+                "exit_destination": "{destination} 方面、{exit}　出口に向かう"
             },
             "left": {
-                "default": "左方向、出口です",
-                "name": "左方向、{way_name}方面、出口です",
-                "destination": "左方向、{destination}方面、出口です",
-                "exit": "左方向、{exit} 出口です",
-                "exit_destination": "左方向、{destination} 方面、{exit}　出口です"
+                "default": "左方向、出口に向かう",
+                "name": "左方向、{way_name}方面、出口に向かう",
+                "destination": "左方向、{destination}方面、出口に向かう",
+                "exit": "左方向、{exit} 出口に向かう",
+                "exit_destination": "左方向、{destination} 方面、{exit}　出口に向かう"
             },
             "right": {
-                "default": "右方向、出口です",
-                "name": "右方向、{way_name}方面、出口です",
-                "destination": "右方向、{destination}方面、出口です",
-                "exit": "右方向、{exit} 出口です",
-                "exit_destination": "右方向、{destination} 方面、{exit}　出口です"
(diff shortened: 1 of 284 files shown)


test/fixtures/v5/off_ramp/left_default.json
```diff
@@ -21,7 +21,7 @@
         "hu": "Hajtson az átkötő útszakaszra balról",
         "id": "Ambil jalan yang melandai di sebelah kiri",
         "it": "Prendi la rampa a sinistra",
-        "ja": "左方向のランプに向かう",
+        "ja": "左方向、出口です",
```
tsuz: "左方向、出口に向かう" is better
"左方向、出口です" means "there is an exit on the left"

languages/translations/ja.json
```diff
@@ -299,90 +299,90 @@
         },
         "off ramp": {
             "default": {
-                "default": "ランプに向かう",
-                "name": "{way_name}に向けてランプに向かう",
-                "destination": "{destination}に向けてランプに向かう",
-                "exit": "{exit}出口に向かう",
-                "exit_destination": "{exit}出口を出て{destination}へ向かう"
+                "default": "出口です",
+                "name": "{way_name}方面、出口です",
```
tsuz: "{way_name}方面、出口に向かう" is better
"{way_name}方面、出口です" means "there is an exit on the {way_name}"

tsuz: I would probably just replace 出口です with 出口に向かう

tichimura: I believe many of Japanese Navigation system will say "出口です” followed with "右側車線を進んでください"(take right lane).  @danpaz we may need to get more user research if you are uncertain for that.

danpaz: @tichimura that is good to know about combining 出口です with lane instructions. This project doesn't support lanes yet, so we could not distinguish between a road with one lane or multiple. At this time it sounds like @tsuz's suggestion to replace all instances of 出口です with 出口に向かう would be the next best option. Do you agree with that?

languages/translations/ja.json
```diff
@@ -299,90 +299,90 @@
...
+                "exit": "右方向、{exit} 出口です",
+                "exit_destination": "右方向、{destination} 方面、{exit}　出口です"
             }
         },
         "on ramp": {
             "default": {
-                "default": "ランプへ向かう",
-                "name": "ランプへ向かい、{way_name}を進む",
-                "destination": "ランプに向かい、その後{destination}へ向かう"
+                "default": "入口です",
```
danpaz: @tsuz would you say the same applies for on ramps? 入口ですへ向かう instead of 入口です?

languages/translations/ja.json
tsuz: @danpaz 

@tichimura has a great point since he has done more extensive comparison with other car navigations. "出口です" seems to be standard among other navs whereas my suggestion was to keep the format standardized among other strings. So, whichever is fine!

languages/translations/ja.json
tsuz: @danpaz 
True, if we change the "on ramps", then this part should also be changed. I guess in that case, we can keep it as "出口です"？

languages/translations/ja.json
danpaz: 👍 thanks both, I kept the original phrasing for off ramp (出口です) and on ramp (入口です).

yuryleb:
But what's about updating strings on Transifex?

PR #291, opened 2019-11-12
Named intersections

danpaz:
# Issue

Branches off #290 as both PRs relate primarily to Japanese guidance.

For routing in Japan and Korea, it's typical to have junctions with names announced in navigation instructions, e.g. "Turn right at Shibuya Crossing" rather than road names. See https://wiki.openstreetmap.org/wiki/Named_spots_instead_of_street_names.

@tsuz @YoshiYazawa @tichimura for review of the Japanese phrases.

## Tasklist

 - [ ] Add changelog entry
 - [ ] Review


index.js
```diff
@@ -190,14 +190,23 @@ module.exports = function(version) {
...
             // Decide which instruction string to use
-            // Destination takes precedence over name
+            // In order of precedence:
+            //   - exit + destination signage
+            //   - destination signage
+            //   - exit signage
+            //   - junction name
+            //   - road name
+            //   - waypoint name (for arrive maneuver)
+            //   - default
```
1ec5: Thanks for updating this comment!

languages/translations/ru.json
```diff
@@ -166,17 +166,20 @@
             "default": {
                 "default": "Поверните {modifier}",
                 "name": "Поверните {modifier} на {way_name:accusative}",
-                "destination": "Поверните {modifier} в направлении {destination}"
+                "destination": "Поверните {modifier} в направлении {destination}",
+                "junction_name": "Turn {modifier} at {junction_name}"
```
1ec5: Based on the key `junction_name`, it sounds like we don’t expect this particular string to be repurposed for destination-based instructions in the future, correct? Since this feature is primarily going to be used in Japan and South Korea, we probably don’t need to worry all that much about cases and declension in languages like Russian, though perhaps grammar rules may eventually be needed for the articles in French.

languages/translations/en.json
```diff
@@ -166,17 +166,20 @@
...
-                "destination": "Continue straight towards {destination}"
+                "destination": "Continue straight towards {destination}",
+                "junction_name": "Continue straight at {junction_name}"
             },
             "uturn": {
                 "default": "Make a U-turn at the end of the road",
                 "name": "Make a U-turn onto {way_name} at the end of the road",
-                "destination": "Make a U-turn towards {destination} at the end of the road"
+                "destination": "Make a U-turn towards {destination} at the end of the road",
+                "junction_name": "Make a U-turn at {junction_name} at the end of the road"
```
1ec5: `end of road` instructions were introduced as a user convenience, so that the user would know they didn’t need to keep track of the remaining distance and could focus on reaching the end of the road. It may be desirable to avoid overloading the user with both the junction name and the “end of road” phrase.

languages/translations/ru.json
yuryleb: Actually this will be already prepared `prepositional` grammar case in Russian 😉 But I'm not sure what changes will be necessary for Hungarian (@kiskunk) and perhaps Danish (@cmahopper)?

kiskunk:
Well for the hungarian version we had to write our own translator engine, since there were way too many grammar rules and problems. So now we use the english version then pass it to our translator java. I have no idea how it could be implemented in OSRM.

languages/translations/en.json
danpaz: Agree, updated in https://github.com/Project-OSRM/osrm-text-instructions/pull/291/commits/a4d24234d91f06295eb882900c334eac34b52f35.

languages/translations/vi.json
```diff
@@ -132,27 +132,32 @@
             "sharp left": {
                 "default": "Quẹo gắt bên trái",
                 "name": "Quẹo gắt bên trái để chạy tiếp trên {way_name}",
-                "destination": "Quẹo gắt bên trái về {destination}"
+                "destination": "Quẹo gắt bên trái về {destination}",
+                "junction_name": "Quẹo gắt bên trái ở {junction_name}"
```
1ec5: 👍 I think the more natural way to say this in Vietnamese would be “Đến {junction_name} quẹo gắt bên trái”. (There are a few famous intersections in Ho Chi Minh City that would get this treatment for sure.) However, in turn-by-turn guidance, it’s more effective to say the maneuver direction before ancillary details like road or junction names, as this translation does.

PR #297, opened 2020-01-15
Japanese translation improvements

danpaz:
# Issue

More follow up from #290 further improving our Japanese translations:
- Remove incorrect / superfluous characters in "nth" arrival instructions, replacing "{nth}目の目的地" ({nth} eye destination") with "{nth}目の目的地" ("{nth} destination").
- Consistently use honorific phrases with -masu suffix.
- Improve non-exit variations of the "off ramp" instruction. Even when there's no exit information we were using "出口" ("exit") in the phrase.

@1ec5 @yuryleb I have not synced these or previous changes with Transifex yet, let me know the best way to go about doing that. Do I simply upload this JSON document in the Transifex UI?

## Tasklist

 - [x] Upate test fixtures
 - [x] Add changelog entry
 - [x] Review

## Requirements / Relations


tsuz:
What actuallu goes into `{nth}`?

danpaz:
> What actually goes into {nth}?

@tsuz Here's the logic that populates the `{nth}` string: https://github.com/Project-OSRM/osrm-text-instructions/blob/master/index.js#L228

Basically:
- If there is only a single waypoint, in a simple A to B route, `{nth}` is replaced by an empty string
- If there are multiple waypoints along a route, `{nth}` is replaced with one of these numbered strings from 1 to 10: https://github.com/Project-OSRM/osrm-text-instructions/blob/master/languages/translations/ja.json#L8-L17

So after the replacement we will have these variations:

n | EN | JA
-- | -- | --
-- | You have arrived at your destination | 目的地に到着しました
1 | You have arrived at your first destination | 1 つ目の目的地に到着しました
2 | You have arrived at your second destination | 2 つ目の目的地に到着しました

danpaz:
Ok, Transifex is now up to date.

1ec5:
Superseded by #312.

PR #301, opened 2021-02-03
Add Persian (Iran) language

Mahdi-Farhani:


akashihi (approved):

akashihi (changes requested):
The content is fine, bu file start with byte order mark which is forbidden by JSOB RFC (Implementations MUST NOT add a byte order mark to the beginning of a JSON text. https://tools.ietf.org/id/draft-ietf-json-rfc4627bis-09.html#rfc.section.8.1) Could you please remove that mark and convert to UTF-8?

akashihi (approved):

akashihi:
@Mahdi-Farhani Looks nice! The only thing i'm concerned - are those version changes in package-lock.json relevant? Looks like you are updating version of the package itself, but you do downgrade some dependencies. Is there any reason for that?

Mahdi-Farhani:
> @Mahdi-Farhani Looks nice! The only thing i'm concerned - are those version changes in package-lock.json relevant? Looks like you are updating version of the package itself, but you do downgrade some dependencies. Is there any reason for that?

Hi, I do not change anything and I don't know why the package-lock is changed. You can Ignore these changes.

akashihi:
@Mahdi-Farhani I'm afraid i can't remove file from the PR, but you can.

Mahdi-Farhani:
> @Mahdi-Farhani I'm afraid i can't remove file from the PR, but you can.

done

akashihi:
@Mahdi-Farhani Sorry for delay :) Well, you are committing file removal, not removing the file from commit. I think you should go back and cancel that change completely. 

I can't guarantee, that it will work for you, but you could try following:

`git rebase -i HEAD~3`

an editor will open. For the last commit, that deletes the file, replace `pick` with `d` and for the middle commit replace `pick` with `e` Save the file then and close the editor.

You should be editing the middle commit now. Please issue the following commands:
    git reset --soft HEAD~1
    git reset HEAD package-lock.json
    git commit -c ORIG_HEAD  
    git rebase --continue

This should remove the 'package-lock.json' file from you PR and then you can do a force push.

yuryleb:
@Mahdi-Farhani, just commit original `package-lock.json` file from master branch and it will be "removed" from your PR.

languages.js
```diff
@@ -86,7 +87,8 @@ var instructions = {
     'uk': instructionsUk,
     'vi': instructionsVi,
     'yo': instructionsYo,
-    'zh-Hans': instructionsZhHans
+    'zh-Hans': instructionsZhHans,
+    'fa-IR': instructionsFaIr
```
yuryleb, line 91: It's better to add your language code in alphabet order not to the the end of the list 😉

languages.js
```diff
@@ -30,6 +30,7 @@ var instructionsUk = require('./languages/translations/uk.json');
 var instructionsVi = require('./languages/translations/vi.json');
 var instructionsYo = require('./languages/translations/yo.json');
 var instructionsZhHans = require('./languages/translations/zh-Hans.json');
+var instructionsFaIr = require('./languages/translations/fa-IR.json');
```
1ec5, line 33: This localization should be under the language code `fa`, not `fa-IR`. We shouldn’t add a localization specific to Iran until we have one that covers the Persian language regardless of the country. Otherwise, clients that don’t specify the country or that specify a different country won’t be able to pick up this localization at all. There may eventually be a need to split out a separate `fa-IR` localization, in order to account for any differences in terminology that are common in Iran but not abroad. However, this would be a premature step for now.

I’m not sure who approved the request to add the [`fa-IR`](https://app.transifex.com/project-osrm/osrm-text-instructions/language/fa_IR/) localization on Transifex, but the translations there should be merged with the localization for [`fa`](https://app.transifex.com/project-osrm/osrm-text-instructions/language/fa/) that was started there. If you need access to that localization in order to complete it, please put in a request to join the translation team. Thanks!

PR #307, opened 2023-01-04
Update fr.json

operon:
In french the roundabout is an "giratoire".
"Giratoire" is circular road with priority to vehicles that turn in circles. junction:roundabout https://osm.org/go/0BM0ndP58-- "Rond-point is a circular road with priority to vehicles entering in circles. junctionj:circular  https://osm.org/go/0BPIBp61A--

### DIFF
--- a/languages/translations/fr.json
+++ b/languages/translations/fr.json
@@ -396,9 +396,9 @@
         "rotary": {
             "default": {
                 "default": {
-                    "default": "Prendre le rond-point",
-                    "name": "Prendre le rond-point, puis sortir sur {way_name:article}",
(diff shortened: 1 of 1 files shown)


yuryleb:
Thanks, but it's better to update translation on https://www.transifex.com/project-osrm/osrm-text-instructions/ first and upload your changes here after (there is some script to do this).

jcoupey:
There is indeed a subtle semantic distinction in french related to priorities as you clearly stated. On the other hand I wonder if it is really worth making a distinction here since it seems to me that most people use both terms as synonyms, and "rond-point" is much more common?

1ec5:
Superseded by #312.

